PR #12, opened 2015-01-04
Added benchmark for select the first element matching in an Enumerable.

nmeylan:
Hi, I figured out that some people use the wrong way to select the first element of an Enumerable that match to _x_.
I wrote a test case that represent the performance issue when using the wrong way.

Bonus : I fixed a test case title in the readme.


JuanitoFatas:
Ohhh thank you so much! Nice work!!! :heart:

PR #13, opened 2015-01-04
Default rake task now perform all test cases under "code/" folder.

nmeylan:
Hello, 
I saw that you are using travis, but when people add test cases they (or you) should append the new test case file into the Rakefile.

Now all test cases are loaded.

Also a test case crashed (explanation in the commit message) and made Travis build error.

Build passed for me : https://travis-ci.org/nmeylan/fast-ruby/builds/45884857

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -1,27 +1,6 @@
 desc 'run benchmark in current ruby'
 task :run_benchmark do
-  [
-    'code/general/assignment.rb',
-    'code/general/begin-rescue-vs-respond-to.rb',
-    'code/proc-and-block/block-vs-to_proc.rb',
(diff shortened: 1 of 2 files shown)


JuanitoFatas:
> I saw that you are using travis, but when people add test cases they (or you) should append the new test case file into the Rakefile.

Ahhh I forget to add them as I merge new things. I want to follow the order as README.md, that's why I did not use ` Dir['code/*/*.rb']`.

> Also a test case crashed (explanation in the commit message) and made Travis build error.

:+1:

JuanitoFatas:
I'll merge first and deal with the order of tests later, thank you very much! :clap: :100:

PR #14, opened 2015-01-17
Add benchmark for #each vs. for loop.

majjoha:
I thought it would be interesting to see how `#each` compares to a regular for loop. So in this pull request I've added the code for the benchmark and an updated README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -320,6 +320,25 @@ Comparison:
      each_with_index:   146266.8 i/s - 1.88x slower
 ```
 
+##### `Enumerable#each` vs `for` loop [code](code/array/each-vs-for-loop.rb)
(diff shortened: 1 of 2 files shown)


JuanitoFatas:
@majjoha Thanks for this pull! Do you think we need to create a specific folder for storing examples for `Enumerable`?

majjoha:
No problem! Actually, there is already a [code/enumerable](https://github.com/JuanitoFatas/fast-ruby/tree/master/code/enumerable) folder. I've moved the benchmark to this folder instead.

JuanitoFatas:
Ohhh thank you I forget there already have one. Awesome work :wink:

PR #15, opened 2015-01-17
Add benchmark for #count vs. #size.

majjoha:
Although, `#count` can be given an argument or take a block, it can also return the size of a collection similar to `#size` as stated in this [blog post](http://batsov.com/articles/2014/02/17/the-elements-of-style-in-ruby-number-13-length-vs-size-vs-count/).

In this pull request, I've added a benchmark for `#count` vs. `#size` and updated the README accordingly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -233,6 +233,23 @@ Comparison:
  Array#shuffle.first:   347632.6 i/s - 17.71x slower
 ```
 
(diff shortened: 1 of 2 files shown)


JuanitoFatas:
:+1: Thank you! Let's collect more examples! :smile:

PR #32, opened 2015-03-06
Array#first vs Array#[](0) and Array#last vs Array#(-1)

vpereira:
first and last are slower then using index. For example in Rubinius, it returns a new array (Array.new[0, n]) while array[idx] returns a Rubinius.primitive :array_new_range


JuanitoFatas:
Thanks! I've committed your Pull Request with some minor fixes. Awesome work :+1:!

PR #33, opened 2015-03-22
assignment for array elements

u2:
Inspired by avoid [run time allocations](https://github.com/wonderbread/rack-pratchett/pull/1)
```
ruby -v code/general/assignment_array_elements.rb
ruby 2.2.1p85 (2015-02-26 revision 49769) [x86_64-darwin14]
element objects Assignment
                        83.252k i/100ms
    array Assignment    81.089k i/100ms
element objects Assignment
                          2.372M (±10.8%) i/s -     11.739M
    array Assignment      2.875M (± 8.2%) i/s -     14.272M

Comparison:
    array Assignment:  2875123.2 i/s
element objects Assignment:  2371948.5 i/s - 1.21x slower

element objects Assignment
                        81.927k i/100ms
    array Assignment    91.828k i/100ms
element objects Assignment
                          2.316M (± 8.8%) i/s -     11.552M
    array Assignment      3.801M (±11.8%) i/s -     18.641M

Comparison:
    array Assignment:  3800505.2 i/s
element objects Assignment:  2316031.8 i/s - 1.64x slower
```

### DIFF
--- a/code/general/assignment_array_elements.rb
+++ b/code/general/assignment_array_elements.rb
@@ -1,0 +1,39 @@
+require 'benchmark/ips'
+
+def slow(env)
+  status, headers, body = env
+  headers.delete('Http-Access-Token')
+  [status, headers, body]
+end
+
+def fast(env)
+  env[1].delete('Http-Access-Token')
+  env
+end
+
+env = [200, { 'Http-Access-Token' => '99d25f556b456d658ba0185540d9daa0' }, 'Hello ips']
+
+Benchmark.ips do |x|
+  x.report('element objects Assignment') { slow(env) }
+  x.report('array Assignment') { fast(env) }
+  x.compare!
+end
+
+def slow_ext(env)
+  status, headers, body = env
+  headers['X-Clacks-Overhead'] = 'GNU Terry Pratchett'
+  [status, headers, body]
+end
+
+def fast_ext(env)
+  env[1]['X-Clacks-Overhead'.freeze] = 'GNU Terry Pratchett'.freeze
+  env
+end
+
+env = [200, { 'Http-Access-Token' => '99d25f556b456d658ba0185540d9daa0', 'X-Clacks-Overhead' => 'Cascade' }, 'Hello ips']
+
+Benchmark.ips do |x|
+  x.report('element objects Assignment') { slow_ext(env) }
+  x.report('array Assignment') { fast_ext(env) }
+  x.compare!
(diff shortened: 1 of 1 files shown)


jonatack:
I get the same results, but it seems like an unfair comparison because the slow version doesn't mutate the supplied array and the fast version does. Thoughts?

jonatack:
It seems to me that a more fair fast version should dup/copy the array first.

jonatack:
My mistake, both the slow and fast versions are mutating directly. Never mind ;)

winston:
Hi! Pardon me, as I am helping @JuanitoFatas to clear the PRs.

I think we would prefer to merge comparison between atomic and generic actions. 
The actions performed in this PR include access and `delete`, and seems to be 
more logic-dependent than generic in nature. So I would lean towards closing this PR.
:kissing_heart:

ixti:
I would also like to point out that comparison is incorrect.
In `slow` version you are creating a new (!) array at the end instead of returning original `env`.
With this patch:

``` diff
--- assignment_array_elements.rb	2016-02-10 21:21:27.403563391 +0100
+++ assignment_array_elements_fixed.rb	2016-02-10 21:22:02.691562471 +0100
@@ -3,7 +3,7 @@
 def slow(env)
   status, headers, body = env
   headers.delete('Http-Access-Token')
-  [status, headers, body]
+  env
 end
 
 def fast(env)
@@ -22,7 +22,7 @@
 def slow_ext(env)
   status, headers, body = env
   headers['X-Clacks-Overhead'] = 'GNU Terry Pratchett'
-  [status, headers, body]
+  env
 end
 
 def fast_ext(env)
```

Results will become:

```
element objects Assignment
                       115.648k i/100ms
    array Assignment   126.859k i/100ms
element objects Assignment
                          4.492M (± 6.2%) i/s -     22.320M
    array Assignment      4.840M (± 0.7%) i/s -     24.230M

Comparison:
    array Assignment:  4840436.0 i/s
element objects Assignment:  4491909.9 i/s - 1.08x slower

element objects Assignment
                       127.003k i/100ms
    array Assignment   131.268k i/100ms
element objects Assignment
                          4.529M (± 3.4%) i/s -     22.607M
    array Assignment      5.262M (± 2.7%) i/s -     26.385M

Comparison:
    array Assignment:  5262202.4 i/s
element objects Assignment:  4529275.0 i/s - 1.16x slower
```

PR #38, opened 2015-04-22
Hash vs OpenStruct

ablignaut:
It's often nicer to reference object instances via properties by using OpenStruct rather than just using a hash. However this shows the price you will pay for it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,6 +95,40 @@ Comparison:
 module_eval with string:     1129.7 i/s - 1.19x slower
 ```
 
+##### `hash` vs `openstruct on access` assuming you already have a hash or an openstruct [code](code/general/hash-vs-openstruct-on-access.rb)
+
+```
+$ ruby -v code/general/hash-vs-openstruct-on-access.rb
+ruby-2.1.3 [ x86_64 ]
+Calculating -------------------------------------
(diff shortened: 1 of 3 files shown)


JuanitoFatas:
Hello @ablignaut, thanks for the Pull Request. This is great! However, I think OpenStruct is [well documented](http://ruby-doc.org/stdlib-2.2.3/libdoc/ostruct/rdoc/OpenStruct.html#class-OpenStruct-label-Implementation-3A):

> This should be a consideration if there is a concern about the performance of the objects that are created, as there is much more overhead in the setting of these properties compared to using a Hash or a Struct.

Maybe submit this one to @bbatsov's [ruby-style-guide](https://github.com/bbatsov/ruby-style-guide) as a guideline?

What do you think? Thank you!

ablignaut:
Hi @JuanitoFatas, Even though it is documented that the performance might be slower, it doesn't tell us how much slower. I thought that is what you are trying to expose by the Fast-Ruby project? We are not saying don't use OpenStruct. We are just pointing out what penalty you will be paying if you decide to use it over hash?

JuanitoFatas:
@ablignaut Thanks! Good to raise the awareness. Thank you for the contribution. :bow:

PR #40, opened 2015-07-01
Three paths for calling method

leosaullv:
there have three paths to call a method. i compare it


JuanitoFatas:
Thanks for your contribution. I've manually merged your commits as 7708a6c :+1:

PR #41, opened 2015-07-05
Add String#sub! vs String#[]=

sshaw:
Note that `String#[]=` raises an `IndexError` if the string does not contain the replacement target 

`s.dup[src]=` is also faster then `s.sub`, was going to add that too but... 

May be nice to add a note to the README about running your rake task (or a single fast-ruby script) against multiple rubies with RVM's `do` command, e.g., `rvm 1.9.3,2.2.1,1.7.19 do rake`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -669,6 +669,34 @@ Comparison:
          String#gsub:   516604.2 i/s - 3.60x slower
 ```
 
+##### `String#sub!` vs `String#gsub!` vs `String#[]=` [code](code/string/sub!-vs-gsub!-vs-[]=.rb)
+```
+$ ruby -v code/string/sub\!-vs-gsub\!-vs-\[\]\=.rb
+ruby 2.2.1p85 (2015-02-26 revision 49769) [x86_64-darwin14]
+Calculating -------------------------------------
+  String#['string']=    51.136k i/100ms
+ String#sub!'string'    38.170k i/100ms
+String#gsub!'string'    27.019k i/100ms
+  String#[/regexp/]=    40.252k i/100ms
+ String#sub!/regexp/    36.228k i/100ms
+String#gsub!/regexp/    21.591k i/100ms
+-------------------------------------------------
+  String#['string']=    899.367k (± 5.4%) i/s -      4.500M
+ String#sub!'string'    547.848k (± 8.5%) i/s -      2.748M
+String#gsub!'string'    347.933k (± 6.1%) i/s -      1.756M
+  String#[/regexp/]=    597.763k (±17.1%) i/s -      2.858M
+ String#sub!/regexp/    414.598k (±28.0%) i/s -      1.884M
+String#gsub!/regexp/    266.487k (± 4.2%) i/s -      1.339M
+
+Comparison:
+  String#['string']=:   899366.7 i/s
+  String#[/regexp/]=:   597762.8 i/s - 1.50x slower
+ String#sub!'string':   547847.6 i/s - 1.64x slower
+ String#sub!/regexp/:   414598.4 i/s - 2.17x slower
+String#gsub!'string':   347933.4 i/s - 2.58x slower
+String#gsub!/regexp/:   266487.2 i/s - 3.37x slower
+```
+
 ##### `attr_accessor` vs `getter and setter` [code](code/general/attr-accessor-vs-getter-and-setter.rb)
 
 > https://www.omniref.com/ruby/2.2.0/files/method.h?#annotation=4081781&line=47
--- a/code/string/sub!-vs-gsub!-vs-[]=.rb
+++ b/code/string/sub!-vs-gsub!-vs-[]=.rb
@@ -1,0 +1,44 @@
+require "benchmark/ips"
+
+URL = "http://www.thelongestlistofthelongeststuffatthelongestdomainnameatlonglast.com/wearejustdoingthistobestupidnowsincethiscangoonforeverandeverandeverbutitstilllookskindaneatinthebrowsereventhoughitsabigwasteoftimeandenergyandhasnorealpointbutwehadtodoitanyways.html"
+
+def fast
+  s = URL.dup
+  s["http://"] = ""
+end
+
+def slow_1
+  s = URL.dup
+  s.sub! "http://", ""
+end
+
+def slow_2
+  s = URL.dup
+  s.gsub! "http://", ""
+end
+
+def slow_3
+  s = URL.dup
+  s[%r{http://}] = ""
+end
+
+def slow_4
+  s = URL.dup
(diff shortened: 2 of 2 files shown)


JuanitoFatas:
Thanks for this :heart_eyes: 

Do you think this could add to this benchmark?

```ruby
s[%r{http://}] = ""
```

Also do you think is worth mentioning in `gsub!` to let people know how slow it is compared to others?

sshaw:
> Do you think this could add to this benchmark (#[] with Regular Expression)?

Good idea. Added.

> Also do you think is worth mentioning gsub! to let people know how slow it is compared to others?

What do you mean, something like "checkout the benchmarks for `gsub!` (with link)" in the `String#[]=` section?

JuanitoFatas:
> What do you mean, something like "checkout the benchmarks for gsub! (with link)" in the String#[]= section?

I think some people will just use `gsub` when they only need `sub` so it is good to include in benchmark to let them have impression that `gsub` is slower than `sub`, also if you don't mind I want to use double quotes :blush: :


```ruby
require "benchmark/ips"

URL = "http://www.thelongestlistofthelongeststuffatthelongestdomainnameatlonglast.com/wearejustdoingthistobestupidnowsincethiscangoonforeverandeverandeverbutitstilllookskindaneatinthebrowsereventhoughitsabigwasteoftimeandenergyandhasnorealpointbutwehadtodoitanyways.html"

def fast
  s = URL.dup
  s["http://"] = ""
end

def slow_1
  s = URL.dup
  s.sub! "http://", ""
end

def slow_2
  s = URL.dup
  s.gsub! "http://", ""
end

def slow_3
  s = URL.dup
  s[%r{http://}] = ""
end

def slow_4
  s = URL.dup
  s.sub! %r{http://}, ""
end

def slow_5
  s = URL.dup
  s.gsub! %r{http://}, ""
end


Benchmark.ips do |x|
  x.report("String#['string']=")   { fast   }
  x.report("String#sub!'string'")  { slow_1 }
  x.report("String#gsub!'string'") { slow_2 }
  x.report("String#[/regexp/]=")   { slow_3 }
  x.report("String#sub!/regexp/")  { slow_4 }
  x.report("String#gsub!/regexp/") { slow_5 }
  x.compare!
end
```

sshaw:
> I think some people will just use gsub when they only need sub so it is good to include in benchmark to let them have impression that gsub is slower than sub,

Yes definitely, I see this a lot :scream_cat: 

> also if you don't mind I want to use double quotes

Sure, I love um. I used singe cause that's what a lot (most) of the code uses. So does [the template in the README](https://github.com/JuanitoFatas/fast-ruby#template).

sshaw:
Updated with `gsub` and double quotes.

JuanitoFatas:
> Sure, I love um. I used singe cause that's what a lot (most) of the code uses. So does the template in the README.

Thanks! Changed :stuck_out_tongue_closed_eyes: 

And thanks for your contribution!!! :kissing_heart:

README.md
```diff
@@ -669,6 +669,34 @@ Comparison:
...
+ String#sub!/regexp/    414.598k (±28.0%) i/s -      1.884M
+String#gsub!/regexp/    266.487k (± 4.2%) i/s -      1.339M
+
+Comparison:
+  String#['string']=:   899366.7 i/s
+  String#[/regexp/]=:   597762.8 i/s - 1.50x slower
+ String#sub!'string':   547847.6 i/s - 1.64x slower
+ String#sub!/regexp/:   414598.4 i/s - 2.17x slower
+String#gsub!'string':   347933.4 i/s - 2.58x slower
+String#gsub!/regexp/:   266487.2 i/s - 3.37x slower
```
gogainda: when I used an example string from the middle of the big url string I got this results
```
String#['string']=:   419706.7 i/s
  String#[/regexp/]=:   357612.7 i/s - 1.17x slower
 String#sub!'string':   337370.1 i/s - 1.24x slower
 String#sub!/regexp/:   315017.3 i/s - 1.33x slower
String#gsub!'string':   221010.8 i/s - 1.90x slower
String#gsub!/regexp/:   185900.8 i/s - 2.26x slower
```

JuanitoFatas: Which Ruby and version you use?

gogainda: ruby 2.2.1p85 (2015-02-26 revision 49769) [i686-linux]
rerun with new version of ruby 
```
String#['string']=:   480872.7 i/s
 String#sub!'string':   363232.0 i/s - 1.32x slower
  String#[/regexp/]=:   325252.5 i/s - 1.48x slower
 String#sub!/regexp/:   285147.2 i/s - 1.69x slower
String#gsub!'string':   222337.3 i/s - 2.16x slower
String#gsub!/regexp/:   160696.7 i/s - 2.99x slower
```

JuanitoFatas: Hmmm what's wrong with your result? :blush:

gogainda: nothing)

mblumtritt:
You should add a remark that the sub-string which you want to replace _must_ exist in the source string when `String#['string']=` is used. This makes this method very different to all others - and not really compareble, or?

sshaw:
> You should add a remark that the sub-string which you want to replace must exist in the source string..
> This makes this method very different to all others - and not really compareble, or?

I pondered this in my [original comment](https://github.com/JuanitoFatas/fast-ruby/pull/41#issue-93139995) and think it's a valid point. Considering that performance may be worse if one has to first check for substring existence or deal with exception handling. 

On the other hand the faster-ruby does not acknowledge the preconditions (or side effects) one considers when writing production code. They just tell you what is faster, for example: `block.call` vs `yield` - should they include `if block.nil?` or `if block_given?`
 
At the very least a comment about `IndexError` should be added to the README or code.  @JuanitoFatas what do you think?

mblumtritt:
Sorry that I missed the original comment. But I would prefer to make it clear for the readers/users if there are different preconditions and side effects. My experience is that there a lot of users out there which will only copy the fastest solution "because that was found in this cool experts banchmark" :flushed: 
@JuanitoFatas: please add such a comment somehow!

BTW: the `yield`/`block` sample is compareable because they need same preconditions and will fail in same cases (but with different exception classes).

JuanitoFatas:
> You should add a remark that the sub-string which you want to replace must exist in the source string when String#['string']= is used. This makes this method very different to all others - and not really compareble, or?

I concur with @mblumtritt, added a note in https://github.com/JuanitoFatas/fast-ruby/commit/1ecd378d535c720be49923e86a5f6de336624eb0.

Thank you for bring this up @mblumtritt, really appreciate it! And thanks @sshaw for helping out.

Cheers. :beers:

mblumtritt:
:+1:

PR #58, opened 2015-08-06
use Array#length instead Array#size

meinac:
Because Array is a block of memory so we can talk about the length of the memory and the right way to get the elements count of it is `length` method.
Also `size` is alias of `length` so the real name of it `length`.


JuanitoFatas:
Thank you very much! Merged as a3ca904.

PR #59, opened 2015-08-06
String#=~ faster than String#match

schneems:
If you're only looking for presence of a match and don't need the matchdata, use String#=~ 

https://github.com/schneems/rails/commit/1bf50badd943e684a56a03392ef0ddafefca0ad7#commitcomment-12572839

### DIFF
--- a/README.md
+++ b/README.md
@@ -682,6 +682,29 @@ Comparison:
            String#=~:   854830.3 i/s - 3.32x slower
 ```
 
+##### `String#match` vs `String#=~` [code ](code/string/match-vs-=~.rb)
+
+> :warning: <br>
+> Sometimes you cant replace `match` with `=~`, <br />
+> This is only useful for cases where you are checkin <br />
+> for a match and not using the resultant match object.
+
+```
+$ ruby -v code/string/match-vs-=~.rb
+ruby 2.2.2p95 (2015-04-13 revision 50295) [x86_64-darwin14]
+Calculating -------------------------------------
+           String#=~    69.889k i/100ms
+        String#match    66.715k i/100ms
+-------------------------------------------------
+           String#=~      1.854M (±12.2%) i/s -      9.155M
+        String#match      1.594M (±11.0%) i/s -      7.939M
+
+Comparison:
+           String#=~:  1853861.7 i/s
+        String#match:  1593971.6 i/s - 1.16x slower
+```
+
+
 ##### `String#gsub` vs `String#sub` [code](code/string/gsub-vs-sub.rb)
 
 ```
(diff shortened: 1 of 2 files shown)


Arcovion:
String#=== is faster

Arcovion:
Regexp#=== is a bit faster

schneems:
Yep

```
           String#=~    66.023k i/100ms
        String#match    60.124k i/100ms
          String#===    70.747k i/100ms
           String#=~      1.798M (±10.3%) i/s -      8.913M
        String#match      1.374M (± 9.7%) i/s -      6.854M
          String#===      1.847M (± 9.8%) i/s -      9.197M
```

Seems good. How should I report this? I would like to have all 3 examples in the code and reported.

Arcovion:
Edited my comment as String#=== was wrong, you want Regexp#===. Added example of how it might look...

JuanitoFatas:
How about list all possibilities?

```ruby
require "benchmark/ips"

def fastest
  "foo".freeze === /boo/
end

def faster
  "foo".freeze =~ /boo/
end

def fast
  /boo/ === "foo".freeze
end

def slow
  "foo".freeze.match(/boo/)
end

Benchmark.ips do |x|
  x.report("String#===") { fastest }
  x.report("Regexp#===") { faster }
  x.report("String#=~") { fast }
  x.report("String#match") { slow }
  x.compare!
end
```

```
ruby -v code/string/match-vs-=\~.rb
ruby 2.2.2p95 (2015-04-13 revision 50295) [x86_64-darwin14]
          String#===   143.440k i/100ms
          Regexp#===    94.853k i/100ms
           String#=~    91.999k i/100ms
        String#match    88.397k i/100ms
          String#===      5.637M (± 7.4%) i/s -     28.114M
          Regexp#===      2.599M (± 5.4%) i/s -     12.995M
           String#=~      2.318M (± 7.7%) i/s -     11.592M
        String#match      1.905M (± 8.4%) i/s -      9.458M

Comparison:
          String#===:  5637360.8 i/s
          Regexp#===:  2599326.4 i/s - 2.17x slower
           String#=~:  2318149.8 i/s - 2.43x slower
        String#match:  1905484.8 i/s - 2.96x slower
```

:question:

Arcovion:
String#=== doesn't work, it's not the correct result.

JuanitoFatas:
> String#=== doesn't work, it's not the correct result.

Yup. The https://github.com/JuanitoFatas/fast-ruby/pull/59#issuecomment-128562623 is wrong.

--

Let's keep this simple since the regexp need to flip the string to be compared to the end. Which is counterintuitive

schneems:
It's a pretty big speed boost, we would have to rename it to `Regexp#===` and we could put a warning in about not switching the arguments cause you'll get the wrong results

PR #60, opened 2015-08-07
Update assignment example to reflect conclusion

sealocal:
If I understand correctly, then the explanation in PR 50 concludes that parallel assignment is faster.  Given that this is the first example called out in the README, I hope this PR would remove doubt or confusion to any Benchmark and/or idiom newbs.


sealocal:
Never mind!  I've seemed to flipped them examples and could have sworn that I saw them the other way around.

PR #61, opened 2015-08-07
name file for end of string matching correctly

Aqualon:


JuanitoFatas:
Thank you @Aqualon :blush:

PR #65, opened 2015-08-11
Explain the Hash#fetch example in more detail

obrok:


JuanitoFatas:
Thank you! @obrok Good work here! :heart:

PR #68, opened 2015-08-17
yield vs block.call misleading

schmurfy:
What is slower is to bind the block to a variable, not calling it, just try my version and you will see that slow and slow4 are both 4.x slower than yield alone.

### DIFF
--- a/README.md
+++ b/README.md
@@ -615,19 +615,20 @@ Comparison:
 ##### `Proc#call` vs `yield` [code](code/proc-and-block/proc-call-vs-yield.rb)
 
 ```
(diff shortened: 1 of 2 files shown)


JuanitoFatas:
Could you also update the README.md? Thanks!

schmurfy:
sure, I will add it today.

michimay:
Can you please contact me asap on 0467515015 regarding all my winnings  and
I'm not sure how this works call me back please asap thankyou or I will
have to unsubscribe

schmurfy:
done.

JuanitoFatas:
@schmurfy Sorry it took so long. Thank you!!!

PR #73, opened 2015-09-13
add comparison of + and concat of Arrays

xjlin0:


Arcovion:
The benchmark is bad, as for `concat` the array is being modified every iteration.
It's also adding itself to itself, meaning the array fills up exponentially.
This doesn't make a difference as it's not actually adding anything, but that in itself makes for a poor test of the methods IMHO.

```ruby
require 'benchmark/ips'

Benchmark.ips do |bm|
  bm.report 'array#concat' do
    array = []
    array.concat ['x']
  end
  bm.report 'array#+' do
    array = []
    array += ['x']
  end
  bm.compare!
end
```
This yields no difference:
```
        array#concat   159.877k i/100ms
             array#+   161.410k i/100ms
        array#concat      4.128M (± 1.4%) i/s -     20.784M
             array#+      4.139M (± 1.8%) i/s -     20.822M

Comparison:
             array#+:  4138811.5 i/s
        array#concat:  4128256.2 i/s - 1.00x slower
```
I think `concat` should be faster, but you need to find a better way of benchmarking it.

xjlin0:
Will it make a better comparison by assign data = [ nil ] instead of [] ?
With [ nil ] #concat is still faster

Arcovion:
For me this benchmark crashes:
   `concat': index 134217728 too big (IndexError)`
The first report adds billions of `nil`s to `data`, then the second report adds those billions together etc.
The reports are order-dependent as `data` is being mutated, so the tests aren't doing what you think they are.

xjlin0:
I change the code to [nil].freeze to use less memory, and hope it won't crash on your environment.  It's true the benchmark is not the correct measurement since it's summary of millions operation, even it shows concat is faster.  How can I make the benchmark valid?  Thanks!

winston:
Hi. Pardon me, I am helping @JuanitoFatas with the PRs.

@xjlin0 are you still working on creating a better benchmark? Thanks!

xjlin0:
No, I cannot think of better way to benchmark it, you can close the PR. However Concat vs + on Array is still an interesting topic.

JuanitoFatas:
If you try this benchmark:

```ruby
# code/array/array-concat-vs-plus.rb
require "benchmark/ips"

def concat
  rdoc_options = ["--markup tomdoc", "--write-options"]
  rdoc_options.concat ["--coverage-report"]
end

def plus
  rdoc_options = ["--markup tomdoc", "--write-options"]
  rdoc_options + ["--coverage-report"]
end

Benchmark.ips do |x|
  x.report("Array#concat") { concat }
  x.report("Array#+")      { plus   }
  x.compare!
end
```

```
$ ruby -v code/array/array-concat-vs-plus.rb
ruby 2.2.3p173 (2015-08-18 revision 51636) [x86_64-darwin14]

        Array#concat    94.840k i/100ms
             Array#+    90.755k i/100ms
        Array#concat      2.283M (± 5.6%) i/s -     11.381M
             Array#+      2.239M (± 6.3%) i/s -     11.254M

Comparison:
        Array#concat:  2282524.7 i/s
             Array#+:  2238532.3 i/s - 1.02x slower
```

They behave roughly the same. So I am going to close this one. Thanks for the Pull Request and @Arcovion reviews. :bow:

PR #74, opened 2015-09-28
Fix idiom title

izumin5210:
I think ... :rabbit: 

* `Hash.fetch` -> `Hash#fetch`
* `Hash#[]` -> `Hash[]` on bracket-vs-dup


JuanitoFatas:
Thanks IZUMI さん! :bear:

PR #76, opened 2015-10-02
Add Function with single Array argument vs splat arguments example

jperville:
Invoking a function that expects multiple splat arguments when all you have is an Array is quite costly (100x). Also, it won't work with sufficiently large arrays as seen [here](http://stackoverflow.com/questions/28703587/systemstackerror-when-pushing-more-than-130798-objects-into-an-array).

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,6 +143,28 @@ Comparison:
              &method:   467095.4 i/s - 4.00x slower
 ```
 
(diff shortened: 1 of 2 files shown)


JuanitoFatas:
Thanks Julien! Nice example :relaxed:

PR #79, opened 2015-10-29
Compare kernel loop and while loop

TheSmartnik:
I thought it is worth sharing as I was always sure that for an infinite loop, `loop` is the way to go


JuanitoFatas:
Thanks @TheSmartnik, did not know while is faster than loop! Nice! :+1:

PR #85, opened 2015-12-17
Set#include is faster than Array#include

nateberkopec:
May want to include a warning that for most other methods (each and other iterators, union/difference), Array is faster.

### DIFF
--- a/README.md
+++ b/README.md
@@ -351,6 +351,27 @@ Comparison:
 
 ```
 $ ruby -v code/array/array-last-vs-index.rb
+ruby 2.2.3p173 (2015-08-18 revision 51636) [x86_64-darwin15]
+
+Calculating -------------------------------------
+        Set#include?    52.470k i/100ms
+  SortedSet#include?    52.462k i/100ms
+      Array#include?    44.830k i/100ms
+-------------------------------------------------
+        Set#include?      1.669M (±11.1%) i/s -      8.238M
+  SortedSet#include?      1.646M (± 8.7%) i/s -      8.184M
+      Array#include?    984.252k (± 9.4%) i/s -      4.931M
+
+Comparison:
+        Set#include?:  1669039.8 i/s
+  SortedSet#include?:  1645635.4 i/s - 1.01x slower
+      Array#include?:   984252.1 i/s - 1.70x slower
+```
+
+##### `Array#include?` vs `Set#include?` [code](code/array/array-include-vs-set-include.rb)
+
+```
+$ ruby -v code/array/array-include-vs-set-include.rb
 ruby 2.2.0p0 (2014-12-25 revision 49005) [x86_64-darwin14]
 
 Calculating -------------------------------------
--- a/code/array/array-include-vs-set-include.rb
+++ b/code/array/array-include-vs-set-include.rb
@@ -1,0 +1,25 @@
+require 'benchmark/ips'
+require 'set'
+
+ARRAY = (1..100).to_a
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -351,6 +351,27 @@ Comparison:
 
 ```
 $ ruby -v code/array/array-last-vs-index.rb
```
ixti: this sould be: ` code/array/array-include-vs-set-include.rb `

nateberkopec: whoops, fixed

ixti:
> May want to include a warning that for most other methods (each and other iterators, union/difference), Array is faster.

I'm not sure about most of the methods. But creation of `Set`s is minimum 5 times slower. That's why union/merge will be 5 times slower as well. So, I guess it worth to mention that, yeah ;))

Arcovion:
Under the hood, a `Set` is implemented as a `Hash`; how about including an example of this too?
```ruby
require 'benchmark/ips'
require 'set'

ARRAY = [*1..1000].shuffle
SET = ARRAY.to_set
SORTED_SET = SortedSet.new(ARRAY)
HASH = ARRAY.each_with_object(Hash.new){ |i, h| h[i] = true }

def fast
  SET.include?(750)
end

def slow
  ARRAY.include?(750)
end

def also_fast
  SORTED_SET.include?(750)
end

def fastest
  HASH.include?(750)
end

Benchmark.ips do |bm|
  bm.report('Hash#include?') { fastest }
  bm.report('Set#include?') { fast }
  bm.report('SortedSet#include?') { also_fast }
  bm.report('Array#include?') { slow }
  bm.compare!
end
```
```
       Hash#include?    86.889k i/100ms
        Set#include?    78.141k i/100ms
  SortedSet#include?   108.497k i/100ms
      Array#include?    25.663k i/100ms
       Hash#include?      8.752M (± 3.0%) i/s -     43.705M
        Set#include?      6.349M (± 2.6%) i/s -     31.725M
  SortedSet#include?      6.385M (± 2.3%) i/s -     31.898M
      Array#include?    290.163k (± 2.6%) i/s -      1.463M

Comparison:
       Hash#include?:  8752135.4 i/s
  SortedSet#include?:  6384910.8 i/s - 1.37x slower
        Set#include?:  6348659.4 i/s - 1.38x slower
      Array#include?:   290162.9 i/s - 30.16x slower
```

nateberkopec:
I think it depends on whether or not you think people will actually be swapping Arrays with Hashes for the `include?` performance. `Set`s are conceptually similar to Arrays, but `Hash`es of course are not. I can't see myself ever using a Hash with all `true` values in place of an Array, that just feels...sinful.

ixti:
@Arcovion is absolutely right. Set is implemented using Hash under the hood. But I don't think we should add `Hash#include?` in here. It's a bit out of scope of this particular case IMO.

ixti:
@Arcovion I think it's better to add info that Set is utilizing Hash internally rather than add `Hash#include?` in here. Because sets can be implemented in more than one way with Hash. ;))

Arcovion:
Depends whether micro-optimizations are ok to be included here, could also add gems to compare against which implement prefix trees or whatnot.

Indeed, it should at least be noted somewhere that there are better algorithms for arrays of simple strings and that `Set` is just a `Hash` with `true` values under the hood. 

There probably needs to be a whole new section for explanations where people can write code snippets, see algorithm examples or gems for people to discover/use, examine what micro-optimisations are available for certain problems etc.

PR #86, opened 2015-12-17
find_index vs index.find

nateberkopec:


code/enumerable/index-find-vs-find-index.rb
```diff
@@ -0,0 +1,17 @@
+require 'benchmark/ips'
+
+ARRAY = (1..100).to_a
+
+def slow
+  ARRAY.index(ARRAY.find { |i| i % 5 == 0 && i % 7 == 0 })
+end
+
+def fast
+  ARRAY.find_index { |i| i % 5 == 0 && i % 7 == 0 }
```
ixti: why not use some fixed number instead?
`%` adds some impact on results, so I believe `ARRAY.find_index { 42 }` might show a bit more correct benchmark numbers

avellable:
```index``` is just an alias of ```find_index```. Let me close this one. :)

reference: [here](https://github.com/ruby/ruby/blob/trunk/array.c#L6116-L6117)

PR #87, opened 2015-12-17
Clarify Proc#call vs yield, add explanation

nateberkopec:


code/proc-and-block/proc-call-vs-yield.rb
```diff
@@ -8,6 +8,10 @@ def slow2 &block
   yield
 end
 
+def slow3 &block
```
ixti: Please wrap argument with brackets.

README.md
```diff
@@ -733,23 +733,29 @@ Comparison:
                Block:    47914.3 i/s - 1.14x slower
 ```
 
-##### `Proc#call` vs `yield` [code](code/proc-and-block/proc-call-vs-yield.rb)
+##### `Proc#call` and block arguments vs `yield`[code](code/proc-and-block/proc-call-vs-yield.rb)
+
+[In MRI Ruby, block arguments are converted to Procs, which incurs a heap allocation](https://www.omniref.com/ruby/2.2.0/symbols/Proc/yield?#annotation=4087638&line=711).
```
ixti: IMO it worth to make link text a bit smaller. :D

``` markdown
In MRI Ruby, block arguments [are converted to Procs](...), which incurs...
```

JuanitoFatas: :+1:

code/proc-and-block/proc-call-vs-yield.rb
JuanitoFatas: :+1:

JuanitoFatas:
@nateberkopec Thank you for the explanations!!! :bow:

PR #88, opened 2015-12-26
Array insert vs unshift

rebelact:
I needed to insert items in the beginning of the array and unshift turns out to be much faster. The comparison here is for a larger array and insert of the beginning of the same array. Initially i did the comparison between `fast` and `slow`, but this does not reflect correctly the difference in the performance.

````ruby
def fast
  [].unshift(1)
end

def slow
  [].insert(0, 1)
end
````

Let me know if further information is required.

### DIFF
--- a/README.md
+++ b/README.md
@@ -365,6 +365,22 @@ Comparison:
           Array#last:  7639254.5 i/s - 1.12x slower
 ```
 
+##### `Array#insert` vs `Array#unshift` [code](code/array/insert-vs-unshift.rb)
+
+```
+$ ruby -v code/array/insert-vs-unshift.rb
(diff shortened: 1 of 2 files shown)


Arcovion:
Thanks, looks good to me!

JuanitoFatas:
:+1:

ibrahima:
I was wondering about this performance difference, so I decided to go spelunking. For anyone interested, my understanding of the C implementation is that some space is left at the front of the array for adding elements at the front: https://github.com/ruby/ruby/commit/fdbd3716781817c840544796d04a7d41b856d9f4

This is as of Ruby 2.0 it seems. Nice to know that we can use this without worrying too much about O(N) costs (unless the array is small, in which case O(N) isn't so bad anyway).

PR #100, opened 2016-02-29
Immutable vs Mutable strings

avellable:
Using 'freeze' increases performances for strings.

```
In Ruby 2.1, "str".freeze is optimized by the compiler to return a single shared frozen strings on every invocation. An alternative "str"f syntax was implemented initially, but later reverted.

Although the external scope of this feature is limited, internally it is used extensively to de-duplicate strings in the VM. Previously, every def method_missing(), the symbol :method_missing and any literal "method_missing" strings in the code-base would all create their own String objects. With Ruby 2.1, only one string is created and shared. Since many strings are commonly re-used in any given code base, this easily adds up. In fact, large applications can expect up to 30% fewer long-lived strings on their heaps in 2.1.

For 2.2, there are plans to expose this feature via a new String#f. There's also a proposal for a magic immutable: string comment to make frozen strings default for a given file.
```

Source: http://tmm1.net/ruby21-fstrings/

### DIFF
--- a/README.md
+++ b/README.md
@@ -938,15 +938,15 @@ Comparison:
 $ ruby -v code/string/mutable_vs_immutable_strings.rb
 ruby 2.2.1p85 (2015-02-26 revision 49769) [x86_64-darwin14]
 Warming up --------------------------------------
-              freeze   111.292k i/100ms
-              normal   106.928k i/100ms
+      Without Freeze   106.961k i/100ms
+         With Freeze   118.210k i/100ms
 Calculating -------------------------------------
-              freeze      5.282M (±14.2%) i/s -     25.931M
-              normal      4.549M (±14.3%) i/s -     22.348M
+      Without Freeze      4.386M (± 6.1%) i/s -     21.927M
+         With Freeze      6.893M (± 6.9%) i/s -     34.281M
 
 Comparison:
-              freeze:  5281661.4 i/s
-              normal:  4548975.9 i/s - same-ish: difference falls within error
+         With Freeze:  6892639.5 i/s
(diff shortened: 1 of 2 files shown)


ixti:
So, you created two `String` objects, and compare how fast `Hash#[]` works with strings... o_O
What really makes sense in terms of immutable strings is:

``` ruby
# allocates new string over and over again
def slow
  "foobar"
end

# keeps and reuses shared string
def fast
  "foobar".freeze
end

Benchmark.ips do |x|
  x.report("slow") { slow }
  x.report("fast") { fast }
  x.compare!
end
```

nateberkopec:
What @ixti said :+1:

avellable:
Yeah, looks like that makes more sense.

```
ruby 2.2.1p85 (2015-02-26 revision 49769) [x86_64-darwin14]
      Without Freeze   106.961k i/100ms
         With Freeze   118.210k i/100ms
      Without Freeze      4.386M (± 6.1%) i/s -     21.927M
         With Freeze      6.893M (± 6.9%) i/s -     34.281M

Comparison:
         With Freeze:  6892639.5 i/s
      Without Freeze:  4386460.4 i/s - 1.57x slower
```

I updated the PR.

JuanitoFatas:
@ixti @nateberkopec Thanks for review 🙇 

@avellable Would you like to be a collaborator :) ?

avellable:
@JuanitoFatas Sure!

JuanitoFatas:
> @JuanitoFatas Sure!

Added, welcome 😃

avellable:
Thank you. :)

PR #102, opened 2016-03-18
Add benchmark for Hash#dig vs #[] vs #fetch

dideler:
Comparison of how Ruby 2.3.0's `Hash#dig` performs to similar methods.

Unsafe retrieval options are included. I can remove them if you consider them noisy.

### DIFF
--- a/code/hash/dig-vs-[]-vs-fetch.rb
+++ b/code/hash/dig-vs-[]-vs-fetch.rb
@@ -11,8 +11,12 @@ Benchmark.ips do |x|
     h[:a][:b][:c][:d][:e]
   end
 
-  x.report 'Hash#[] fallback' do
+  x.report 'Hash#[] ||' do
     ((((h[:a] || {})[:b] || {})[:c] || {})[:d] || {})[:e]
+  end
+
+  x.report 'Hash#[] &&' do
+    h[:a] && h[:a][:b] && h[:a][:b][:c] && h[:a][:b][:c][:d] && h[:a][:b][:c][:d][:e]
   end
 
   x.report 'Hash#fetch' do
--- a/code/hash/dig-vs-[]-vs-fetch.rb
+++ b/code/hash/dig-vs-[]-vs-fetch.rb
@@ -1,0 +1,27 @@
+require 'benchmark/ips'
+
+h = { a: { b: { c: { d: { e: "foo" } } } } }
+
+Benchmark.ips do |x|
+  x.report 'Hash#dig' do
+    h.dig(:a, :b, :c, :d, :e)
+  end
+
+  x.report 'Hash#[]' do
+    h[:a][:b][:c][:d][:e]
+  end
+
+  x.report 'Hash#[] fallback' do
+    ((((h[:a] || {})[:b] || {})[:c] || {})[:d] || {})[:e]
+  end
+
+  x.report 'Hash#fetch' do
+    h.fetch(:a).fetch(:b).fetch(:c).fetch(:d).fetch(:e)
+  end
+
+  x.report 'Hash#fetch fallback' do
+    h.fetch(:a, {}).fetch(:b, {}).fetch(:c, {}).fetch(:d, {}).fetch(:e, nil)
+  end
+
+  x.compare!
+end
--- a/README.md
+++ b/README.md
@@ -573,6 +573,32 @@ Comparison:
   Hash#fetch, string:  3981166.5 i/s - 1.89x slower
 ```
 
+##### `Hash#dig` vs `Hash#[]` vs `Hash#fetch` [code](code/hash/dig-vs-[]-fetch.rb)
+
+```
+$ ruby -v code/hash/dig-vs-\[\]-vs-fetch.rb
+ruby 2.3.0p0 (2015-12-25 revision 53290) [x86_64-darwin15]
+Warming up --------------------------------------
+            Hash#dig   144.192k i/100ms
+             Hash#[]   148.853k i/100ms
+    Hash#[] fallback   149.761k i/100ms
+          Hash#fetch   132.257k i/100ms
+ Hash#fetch fallback   120.420k i/100ms
+Calculating -------------------------------------
+            Hash#dig      6.253M (± 5.9%) i/s -     31.145M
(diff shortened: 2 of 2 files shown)


code/hash/dig-vs-[]-vs-fetch.rb
```diff
@@ -0,0 +1,27 @@
...
+  x.report 'Hash#dig' do
+    h.dig(:a, :b, :c, :d, :e)
+  end
+
+  x.report 'Hash#[]' do
+    h[:a][:b][:c][:d][:e]
+  end
+
+  x.report 'Hash#[] fallback' do
+    ((((h[:a] || {})[:b] || {})[:c] || {})[:d] || {})[:e]
```
Comment (nateberkopec) [inline code/hash/dig-vs-[]-vs-fetch.rb]: ouch, do people ever really do it this way? far more common I think is:

```ruby
h[:a] && h[:a][:b] && ## etc
```

Comment (dideler) [inline code/hash/dig-vs-[]-vs-fetch.rb]: I'm not sure which way is more common. Should I use one example over the other or include both?

Comment (deniskorobicyn) [inline code/hash/dig-vs-[]-vs-fetch.rb]: Actually, I think, the way proposed by @nateberkopec is more effective in case when none of the keys are exists, because you don't need to create all of these hashes.

Comment (dideler) [inline code/hash/dig-vs-[]-vs-fetch.rb]: I've added the more common style to the benchmark, see 95b8ea491d5c4ab310c1166c2a280cd7f6356272

code/hash/dig-vs-[]-vs-fetch.rb
```diff
@@ -0,0 +1,31 @@
...
+  x.report 'Hash#[] &&' do
+    h[:a] && h[:a][:b] && h[:a][:b][:c] && h[:a][:b][:c][:d] && h[:a][:b][:c][:d][:e]
+  end
+
+  x.report 'Hash#fetch' do
+    h.fetch(:a).fetch(:b).fetch(:c).fetch(:d).fetch(:e)
+  end
+
+  x.report 'Hash#fetch fallback' do
+    h.fetch(:a, {}).fetch(:b, {}).fetch(:c, {}).fetch(:d, {}).fetch(:e, nil)
```
Comment (ixti) [inline code/hash/dig-vs-[]-vs-fetch.rb]: `fetch(:a, {})` creates new object EVERY time even when it's not needed.
More appropriate would be:

``` ruby
h.fetch(:a) { {} }.fetch(:b) { {} }...
```

ixti:
More interesting would be to compare `dig` vs alternatives upon "broken" path:

``` ruby
H = { :a => { :b => { :d => true } } }

H.dig(:a, :b, :c, :d) # => nil
H[:a] && H[:a][:b] && H[:a][:b][:c] && H[:a][:b][:c][:d] # => nil
```

code/hash/dig-vs-[]-vs-fetch.rb
Comment (mblumtritt) [inline code/hash/dig-vs-[]-vs-fetch.rb]: Your notation will create Procs for every #fetch. That maybe faster or not. Seems that's another test… :)

Comment (dideler) [inline code/hash/dig-vs-[]-vs-fetch.rb]: @ixti I'd like to keep it as is because I see the current approach more often.

dideler:
> More interesting would be to compare dig vs alternatives upon "broken" path:

I find the safe path most interesting because you can compare more options, but here are all cases.

**Safe:** `h = { a: { b: { c: { d: { e: "foo" } } } } }`
```
Comparison:
             Hash#[]:  6676415.9 i/s
            Hash#dig:  6215966.7 i/s - same-ish: difference falls within error
          Hash#[] ||:  6160177.6 i/s - same-ish: difference falls within error
          Hash#fetch:  4424551.0 i/s - 1.51x slower
 Hash#fetch fallback:  3278599.3 i/s - 2.04x slower
          Hash#[] &&:  3096090.4 i/s - 2.16x slower
```

**Broken at last key:** `h = { a: { b: { c: { d: { x: "foo" } } } } }`
The `Hash#fetch` benchmark is excluded because it doesn't gracefully fail.
```
Comparison:
             Hash#[]:  6391117.6 i/s
          Hash#[] ||:  5775857.7 i/s - same-ish: difference falls within error
            Hash#dig:  5599916.6 i/s - same-ish: difference falls within error
 Hash#fetch fallback:  3278820.7 i/s - 1.95x slower
          Hash#[] &&:  2920215.9 i/s - 2.19x slower
```

**Broken at intermediate key:** `h = { a: { b: { c: { x: { e: "foo" } } } } }`
The `Hash#fetch` and `Hash#[]` benchmarks are excluded because they do not gracefully fail.
```
Comparison:
            Hash#dig:  5525802.5 i/s
          Hash#[] ||:  4866123.4 i/s - same-ish: difference falls within error
          Hash#[] &&:  3727126.8 i/s - 1.48x slower
 Hash#fetch fallback:  2921706.4 i/s - 1.89x slower
```

`Hash#dig` does well in all cases. It reads better than the other options and is consistently performant.

ixti:
@dideler I'm absolutely agree that `dig` is a winner :D and I guess the only my proposal is to add link to https://github.com/JuanitoFatas/fast-ruby/pull/102#issuecomment-198827506 in readme :D

code/hash/dig-vs-[]-vs-fetch.rb
Comment (ixti) [inline code/hash/dig-vs-[]-vs-fetch.rb]: @mblumtritt in fact it will not. Why would it create Proc?

In any case, @dideler I tend to agree that probably some people use `fetch(key, {})`.
And that looks pretty fine, but in this case it worth to "cache" that object:

``` ruby
o = {}
h.fetch(:a, o).fetch(:b, o)
```

That's just my thoughts. I don't insist on anything :D

dideler:
:+1:, how's 437d8f86e671bb0357d63fd5dae6dd25ab3f1bf2?

ixti:
LGTM! :+1:

JuanitoFatas:
Thanks everyone 👏 ! Merged as d6e1ac0.

@dideler Dennis would you like to be a collaborator to this repository 😊 ?

dideler:
Thanks @JuanitoFatas - I would love to be a collaborator!

JuanitoFatas:
> Thanks @JuanitoFatas - I would love to be a collaborator!

Yay! ❤️

PR #103, opened 2016-03-21
remove an accidentally duplicated sentence

rousisk:
I think that 414f86f6 introduced the same sentence (regarding `Exception2MessageMapper`) twice so I removed the latter reference (accidentally placed within a code sample).


ixti:
:+1:  Thanks!

PR #105, opened 2016-04-04
compare inject block vs manual entries

kbrock:
Someone had asked me about using the block short form vs manually entering the code.

Turns out, in 1.8.7, using the block was 1.7x slower, but since then, the symbol form is a little quicker.

Thanks for this repo

### DIFF
--- a/code/enumerable/inject-symbol-vs-block.rb
+++ b/code/enumerable/inject-symbol-vs-block.rb
@@ -1,0 +1,24 @@
+require "rubygems"
+require "benchmark/ips"
+
+ARRAY = (1..1000).to_a
+
+def fastest
+  ARRAY.inject(:+)
+end
+
+def fast
+  ARRAY.inject(&:+)
+end
+
+def slow
+  ARRAY.inject { |a, i| a + i }
+end
+
+Benchmark.ips do |x|
+  x.report('inject symbol') { fastest }
+  x.report('inject to_proc') { fast }
+  x.report('inject block')   { slow }
+
+  x.compare!
+end
--- a/README.md
+++ b/README.md
@@ -540,6 +540,27 @@ Enumerable#sort_by (Symbol#to_proc):    25916.1 i/s
      Enumerable#sort:                   14018.3 i/s - 1.85x slower
 ```
 
+##### `Enumerable#inject Symbol` vs `Enumerable#inject Proc` [code](code/enumerable/inject-symbol-vs-block.rb)
+
+Of note, `to_proc` for 1.8.7 is considerable slower than the block format
+
+```
+$ ruby -v code/enumerable/inject-sum-vs-block.rb 
+ruby 2.2.4p230 (2015-12-16 revision 53155) [x86_64-darwin14]
+Warming up --------------------------------------
+       inject symbol     1.893k i/100ms
+      inject to_proc     1.583k i/100ms
+        inject block     1.390k i/100ms
+Calculating -------------------------------------
+       inject symbol     19.001k (± 3.8%) i/s -     96.543k
+      inject to_proc     15.958k (± 3.5%) i/s -     80.733k
+        inject block     14.063k (± 3.9%) i/s -     70.890k
+
+Comparison:
+       inject symbol:    19001.5 i/s
+      inject to_proc:    15958.3 i/s - 1.19x slower
+        inject block:    14063.1 i/s - 1.35x slower
+```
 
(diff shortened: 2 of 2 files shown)


code/enumerable/inject-sum-vs-block.rb
```diff
@@ -0,0 +1,19 @@
+require "benchmark/ips"
+
+User  = Struct.new(:name)
```
Arcovion: Unused variable
How about `ARRAY.inject(:+)` too?

code/enumerable/inject-sum-vs-block.rb
```diff
@@ -0,0 +1,19 @@
...
+ARRAY = (1..10_000_000).to_a
+
+def fast
+  ARRAY.inject(&:+)
+end
+
+def slow
+  ARRAY.inject(0) { |a, i| a + i }
```
Arcovion: This should be `ARRAY.inject { |a, i| a + i }`, or the other should be `ARRAY.inject(0, :+)`

nateberkopec:
[Already covered by Symbol#to_proc IMO](https://github.com/JuanitoFatas/fast-ruby#block-vs-symbolto_proc-code)

nateberkopec:
Whoops, didnt mean to unilaterally close it. My bad!

code/enumerable/inject-sum-vs-block.rb
kbrock: lol. must have been a copy paste from another enumerable one. oops

kbrock:
Made fixes.

Think @nateberkopec is right. that is covered.
Sorry I didn't see it.

agreed to close?

Arcovion:
Just tested, seems like even before with `0` given the block ones were still exactly equal. The symbol one is much faster:
```ruby
require "benchmark/ips"

ARRAY = (1..10_000_000).to_a

def fastest
  ARRAY.inject(:+)
end

def fast
  ARRAY.inject(&:+)
end

def slow
  ARRAY.inject { |a, i| a + i }
end

Benchmark.ips do |x|
  x.report('inject symbol') { fastest }
  x.report('inject block') { fast }
  x.report('inject sum')   { slow }

  x.compare!
end
```
```
       inject symbol     1.000  i/100ms
        inject block     1.000  i/100ms
          inject sum     1.000  i/100ms
       inject symbol      2.372  (± 0.0%) i/s -     12.000 
        inject block      1.668  (± 0.0%) i/s -      9.000  in   5.398692s
          inject sum      1.667  (± 0.0%) i/s -      9.000  in   5.408312s

Comparison:
       inject symbol:        2.4 i/s
        inject block:        1.7 i/s - 1.42x slower
          inject sum:        1.7 i/s - 1.42x slower
```
So I think it's worth having a benchmark to show this even though it's very specific and only applies to `#inject`/`#reduce`.
You need to add the benchmark results to the README (I think it was there before, but not in your last commit?)

kbrock:
aah - symbol. forgot that one. Yea. there was a reason why I was visiting this code :)

kbrock:
@Arcovion what version of ruby are you running?
I've tried with small arrays and larger arrays, and they never show sum and block with the same values

Arcovion:
`ruby 2.3.0p0 (2015-12-25 revision 53290) [x86_64-linux]`
Doesn't matter what Ruby you use for the benchmark, no problem if the results differ.

kbrock:
@Arcovion @nateberkopec Is this good for you? Any other changes?

Arcovion:
Thanks, LGTM.

code/enumerable/inject-symbol-vs-block.rb
```diff
@@ -0,0 +1,24 @@
+require "rubygems"
+require "benchmark/ips"
+
+ARRAY = (1..1000).to_a
+
+def fastest
```
nirvdrum: I think naming the methods after their relative speed is a bit presumptuous. They're subject to change across runtimes and versions. I think it'd be better to name them after the type of operation being performed.

kbrock: +1. I was trying to follow the style guide but I agree

nirvdrum: Yeah, sorry. I see now that that could be read as an attack. I just meant the convention in general should change, IMHO.

kbrock: lol. we're good.

PR #106, opened 2016-04-26
little fix for the hash merge comparison

lvl0nax:
'fast' method has different result from other methods. Because:
```ruby
ORIGINAL_HASH = { foo: "foo" }
#fast
{ foo: 'bar' }.merge!(ORIGINAL_HASH)
#=> {:foo=>"foo"}

#slow
ORIGINAL_HASH.merge(foo: 'bar')
#=> {:foo=>"bar"}

#slow_dup
ORIGINAL_HASH.dup.merge!(foo: 'bar')
#=> {:foo=>"bar"}

#new fast
{ foo: 'bar' }.merge!(ORIGINAL_HASH){ |_key, left, _right| left }
#=> {:foo=>"bar"}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -707,21 +707,21 @@ Comparison:
 
 ```
 $ ruby -v code/hash/merge-bang-vs-merge-vs-dup-merge-bang.rb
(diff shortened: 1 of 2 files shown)


Arcovion:
Thanks!

PR #111, opened 2016-05-12
Update gsub-vs-tr.rb

equivalent:
took me a while to figure out that that's not part of core  `ruby` `benchmark` lib


equivalent:
Ahh ok I thought this is `gits` didn't know this is project sorry closing

Issue #110, opened 2016-05-09
Move away from comparative speed method names


nirvdrum:
This is a proposal to move away from method names like `slow`, `fast`, and `fastest` and instead use something named after the operation. The convention is a bit error prone when authoring the benchmark to begin with, since you can't know a priori what the relative speeds are until you've written the benchmarks. Thus, it requires a renaming of the methods after the benchmark is run. It may introduce confirmation bias as well.

The real problem, I think, is it risks the entire repository becoming bit rotted. Ruby's performance is not fixed in time. If Ruby 3 manages to hit its 3x performance gain, it may very well come at the cost of reshuffling the relative performance of methods involved in these benchmarks. Then we'd be looking at the unenviable case of the `fast` method being faster than the `fastest` method and so on.

Going along with performance being implementation-specific across MRI versions, the problem is exacerbated when running the benchmarks on alternative Ruby implementations. In that case, the relative method values do not hold in many cases.

nateberkopec:
I don't think anyone would be against a PR that updated/changed all of the current benchmarks in the way that you propose. 👍

etagwerker:
Sounds good, we can continue the discussion in #215

Issue #96, opened 2016-01-25
Some benchmarks are measured with wrong methodology


radarek:
### How to measure atom with a ruler?

**tl;dr** When we measure small things we have to remove all overheads which have impact. Wrapping with a method and block call is quite big overhead for most of benchmarks in this repository.

Suppose you want compare performance of "2 + 3" vs "2 * 3" calls. If you do it with a methodology used in this repository you will get this results:

```ruby
require "benchmark/ips"

def slow
  2 * 2
end

def fast
  2 + 2
end

Benchmark.ips do |x|
  x.report("2 * 2") { slow }
  x.report("2 + 2") { fast }
  x.compare!
end
```

(simplified output)
```
Comparison:
               2 + 2:  8304760.0 i/s
               2 * 2:  7535516.6 i/s - 1.10x slower
```

But there is one problem. Calling a method + surrounding block (`{ slow }`) has bigger overhead than calling `Fixnum#+` or `Fixnum#*` itself. Thea easiest way to observe it is to repeat benchmarked operations in one call. Like this:

```ruby
$ cat add_vs_mul2.rb
require "benchmark/ips"

def slow
  2*2; 2*2; 2*2; 2*2; 2*2; 2*2; 2*2; 2*2; 2*2; 2*2;
end

def fast
  2+2; 2+2; 2+2; 2+2; 2+2; 2+2; 2+2; 2+2; 2+2; 2+2;
end

Benchmark.ips do |x|
  x.report("2 * 2") { slow }
  x.report("2 + 2") { fast }
  x.compare!
end
```

```
Comparison:
               2 + 2:  4680545.3 i/s
               2 * 2:  3468681.3 i/s - 1.35x slower
```

See how results changed? Writing our benchmarks in this way would be quite problematic. Fortunately benchmark-ips gem has answer for that. `Benchmark::IPS::Job#report` method allows to pass string which will be compiled before benchmark is run. Passing right string allows to measure it properly:

```ruby
require "benchmark/ips"

Benchmark.ips do |x|
  x.report("2 * 2", "2 * 2;" * 1_000)
  x.report("2 + 2", "2 + 2;" * 1_000)
  x.compare!
end
```

```
Comparison:
               2 + 2:    91567.5 i/s
               2 * 2:    57994.4 i/s - 1.58x slower
```

[...]

PR #112, opened 2016-05-24
Add String#start_with? vs String#[].==

jonathanhefner:
Compares e.g. `url.start_with?("http")` vs `url[0...4] == "http"`

A lot of the existing string benchmarks list their "slow" case first, but the contributing guidelines said to list the "fast" case first.  If I need to change anything, I would be happy to.  Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -884,6 +884,24 @@ Comparison:
            String#=~:   891124.1 i/s - 3.30x slower
 ```
 
+##### `String#start_with?` vs `String#[].==` [code](code/string/start_with-vs-substring-==.rb)
+
+```
+ruby 2.2.4p230 (2015-12-16 revision 53155) [i386-cygwin]
+
+Calculating -------------------------------------
+  String#start_with?    739.364k (± 0.5%) i/s -      3.720M in   5.031116s
+    String#[0, n] ==    346.112k (± 0.2%) i/s -      1.753M in   5.063586s
+   String#[RANGE] ==    318.284k (± 0.2%) i/s -      1.612M in   5.066202s
+   String#[0...n] ==    220.535k (± 0.1%) i/s -      1.110M in   5.033769s
+
+Comparison:
+  String#start_with?:   739364.0 i/s
+    String#[0, n] ==:   346112.1 i/s - 2.14x slower
+   String#[RANGE] ==:   318283.6 i/s - 2.32x slower
+   String#[0...n] ==:   220534.9 i/s - 3.35x slower
+```
+
 ##### `Regexp#===` vs `String#match` vs `String#=~` [code ](code/string/===-vs-=~-vs-match.rb)
 
 > :warning: <br>
--- a/code/string/start_with-vs-substring-==.rb
+++ b/code/string/start_with-vs-substring-==.rb
@@ -1,0 +1,34 @@
+require 'benchmark/ips'
+
+PREFIX = '_'
+STRINGS = (0..9).map{|n| "#{PREFIX if n.odd?}#{n}" }
+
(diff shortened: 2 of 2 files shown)


jonathanhefner:
Based on the conversations in #96, I tried a different approach to these benchmarks:

```ruby
require 'benchmark/ips'

PREFIX = '_'
STRINGS = (0..99).map{|n| "#{PREFIX if n.odd?}#{n}" }

FAST = STRINGS.each_index.map do |i|
  "STRINGS[#{i}].start_with?(PREFIX)"
end.join(';')

SLOW_LENGTH = STRINGS.each_index.map do |i|
  # use `eql?` instead of `==` to prevent warnings
  "STRINGS[#{i}][0, PREFIX.length].eql?(PREFIX)"
end.join(';')

SLOW_RANGE = STRINGS.each_index.map do |i|
  # use `eql?` instead of `==` to prevent warnings
  "STRINGS[#{i}][0...PREFIX.length].eql?(PREFIX)"
end.join(';')

Benchmark.ips do |x|
  x.report('String#start_with?', FAST)
  x.report('String#[0, n] ==', SLOW_LENGTH)
  x.report('String#[0...n] ==', SLOW_RANGE)
  x.compare!
end
```

The code is less clear, but the results show a sharper contrast:
```
$ ruby -v code/string/start_with-vs-substring-\=\=.rb
ruby 2.2.4p230 (2015-12-16 revision 53155) [i386-cygwin]

  String#start_with?     76.406k (± 0.3%) i/s -    387.296k in   5.068975s
    String#[0, n] ==     33.604k (± 0.2%) i/s -    169.364k in   5.039945s
   String#[0...n] ==     21.865k (± 0.3%) i/s -    111.176k in   5.084763s

Comparison:
  String#start_with?:    76405.7 i/s
    String#[0, n] ==:    33604.4 i/s - 2.27x slower
   String#[0...n] ==:    21864.8 i/s - 3.49x slower
```

If desired, I can update this PR using the above code.  Thoughts?

nirvdrum:
While I guess it's a question of what exactly you're trying to benchmark, note that in the `SLOW_LENGTH` case you're performing an extra string length calculation on every call and in the `SLOW_RANGE` case you're making that same extra string length calculation and constructing a range.

The code as written is probably representative of what most people would do in the real world. But if your focus is on measuring the cost of the various prefix checks themselves, it'd probably be worthwhile to calculate the values once outside the loop and then substitute them in.

jonathanhefner:
> The code as written is probably representative of what most people would do in the real world.

I'm glad you see it as such--that's exactly what I intended. :)  Specifically I've seen this pattern a few times:
```ruby
if template_name.first == "_"
  # handle partial view template
end
```

This uses Active Support's [`String#first`](http://apidock.com/rails/ActiveSupport/CoreExtensions/String/Access/first) method, which is implemented just like `slow_range`, including the length retrieval and range allocation.

jonathanhefner:
I've modified the commit to include the changes discussed above.  I've included a case with a pre-allocated range, as per @nirvdrum's suggestion.  I've also renamed the cases as per the discussion in #110.  Thank you for the feedback. :)

JuanitoFatas:
@nirvdrum Thank you for the review!
@jonathanhefner Thanks for your awesome work! 

@nirvdrum @jonathanhefner May I add you as collaborators to this repository?

jonathanhefner:
@JuanitoFatas Sure thing!  Thanks very much! :smiley: :tada:

nirvdrum:
@JuanitoFatas That'd be fantastic. Thanks!

PR #113, opened 2016-06-19
Benchmark presence check

TheSmartnik:
I've been wondering whether the new safe method navigation syntax is fast. Turned out it's almost as fast as simple method invocation.


ixti:
You are testing happy path, where calling method itself is really fastest way (no surprises here).
What sad man operator actually gives is that you avoid chains like `o && o.foo && o.foo.bar`.
That provides you with:

a) slimmer and cleaner code `a&.foo&.bar`
b) handles situation when middle call (`#foo` in the above) returns nil
c) memoizes middle calls

by *c* i mean, that `a&.foo&.bar` is actually a shorthand syntax for:

``` ruby
(_x = a) && (_x = _x.foo) && _x.bar
```

Now,  imagine following, you have `Book` class and `User` that can borrow it in Library. Obviously in this case `Book` might be or might not be borrowed by user:

``` ruby
class Book
  # @return [User, nil]
  def borrower
    # ...
  end
```

Now here's example where `&.` comes into play:

``` ruby
if book&.borrower&.student?
  # ...
else
  # book is either NOT borrowed or borrower user is not student
end
```

Obviously example is pretty stupid, but still demonstrates WHY `&.` is needed and WHY it's invalid to compare it to simple method invocation. When `nil` in the middle of chain is an exception - then you MUST use simple method invocation. When middle call evaluation MIGHT return nil as a valid response you should use `&.`. All simple :D

So, in other words the only comparison that makes sens here is:

``` ruby
a &. b &. c

# vs

a && a.b && a.c

# vs

a.try!(:b).try!(:c)
```

Notice that `&.` acts like Rails' `#try!` not `#try`. It will raise `NoMethodError` if intermediate object is not nil and does not respond to given method.

Gemfile
```diff
@@ -1,5 +1,5 @@
 source 'https://rubygems.org'
 
 gem 'benchmark-ips', '>= 2.0'
-
+gem 'activesupport'
```
nirvdrum: I don't think we should have a dependency on ActiveSupport. And if we do, it really should be locked down, otherwise the results are subject to change every time Rails issues a release of any sort.

TheSmartnik:
@ixti simple method calling here is present in this benchmark just to showcase how fast safe navigation operator is. Maybe I should've make it more clear by adding note to readme instead of adding it to benchmark

Gemfile
TheSmartnik: @nirvdrum Agree.
As another solution, I can just add implementation of `try!` in benchmark file

ixti:
> Maybe I should've make it more clear by adding note to readme instead of adding it to benchmark.

Yeah. It's better to add that safe navigation with happy path is nearly as fast as simple calls chain.

Gemfile
ixti: I think it's fine to lock down to the latest stable as of now, and add `:require => false`.

code/general/safe-navigation-vs-explicit-check.rb
```diff
@@ -0,0 +1,26 @@
+require "benchmark/ips"
+require 'active_support'
+require 'active_support/core_ext/object/try'
```
ixti: Use double quotes here, please

code/general/safe-navigation-vs-explicit-check.rb
```diff
@@ -0,0 +1,26 @@
...
+
+def slow
+  o = Object.new
+  o && o.itself && o.itself.itself
+end
+
+def slowest
+  o = Object.new
+  o.try(:itself).try(:itself)
+end
```
ixti: It's better to create object once outside of benchmarks:

``` ruby
OBJ = Object.new

def fast
  OBJ &.itself &.itself
end
```

TheSmartnik:
@ixti thanks for your feedback. Fixed.

code/general/safe-navigation-vs-explicit-check.rb
```diff
@@ -0,0 +1,24 @@
...
+def fast
+  OBJ &.itself &.itself
+end
+
+def slow
+  OBJ && OBJ.itself && OBJ.itself.itself
+end
+
+def slowest
+  OBJ.try(:itself).try(:itself)
```
deniskorobicyn: Did you meant `try!` here ?

TheSmartnik: @deniskorobicyn yeah, thanks 👍

etagwerker (changes requested):
@TheSmartnik Thanks for sending this over and apologies for the delay! 

Could you please fix the conflicts and update the `activesupport` dependency? 

Thank you!

etagwerker:
Closing this until someone addresses the issues that were mentioned in this PR.

PR #115, opened 2016-07-03
Add Struct to the Hash vs OpenStruct bench

guizmaii:
In the CI results, we can see that since Ruby 2.2 Struct was improved a lot !
We can also see that in JRuby 9 Hashes are better than Structs. Fun :)


code/general/struct-vs-hash-vs-openstruct.rb
```diff
@@ -0,0 +1,40 @@
+require "benchmark/ips"
+require "ostruct"
+
+
+Example = Struct.new(:field_1, :field_2)
+Struct.new('User', :field_1, :field_2)
```
nateberkopec: Is there a reason to do this instead of:

```
User = Struct.new(:field_1, :field_2)
```

guizmaii: Yes ! to compare ! 
I saw the second form in the Ruby doc : http://ruby-doc.org/core-2.2.0/Struct.html

PR #116, opened 2016-07-07
Add Hash#[] vs Hash#include?

chen7897499:


ixti:
First of all, `#include?` (or `#key?`) and `#[]` are doing completely different. Thus they are not interchangeable:

``` ruby
H = { :a => true, :b => false, :c => nil }

H.keys.map { |key| :ok if H.include?(key) } # => [:ok, :ok, :ok]
H.keys.map { |key| :ok if H[key] } # => [:ok, nil, nil]
```

And I'm pretty curious why:

``` ruby
x.report do |n|
  i = 0
  while i < n
    ...
    i += 1
  end
end
```

`Benchmark.ips` already taking care of multiple runs of a block. If you need more cycles, just configure it. In this particular case you are adding strange (and IMHO useless) overhead.

chen7897499:
@ixti  could you show me a example of configure cycles? thank you

PR #119, opened 2017-01-04
ADD: example for String#sub vs String#slice!

roxer:


ixti:
There's no point in comparing bang methods VS non-bang. As `slice!` changes string in-place.
Also comparing sub to slice is kinda strange. `sub` replaces substring matching regexp with the other string, and slice instead returns matching substring

jonathanhefner:
As @ixti said, sub vs slice is a bit odd.  It would be helpful to know what end result you're trying to achieve before deciding which methods to compare.  Also, your fastest test case is labeled as `String#[integer, integer]` but it is really `String#[Range]` (the two are different in terms of semantics).

If you do want to compare bang methods (which involve mutation), I wrote a gem that can help with that: [benchmark-inputs](https://github.com/jonathanhefner/benchmark-inputs).  You can use it just like benchmark-ips, and it will provide similar output.  Let me know if you have any questions.

PR #121, opened 2017-03-18
compare gsub vs tr vs delete to remove character from string

mojavelinux:
This benchmark proves that delete is faster than tr and both are considerably faster than gsub when removing a character from a string.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1329,6 +1329,24 @@ Comparison:
          String#gsub:   516604.2 i/s - 3.60x slower
 ```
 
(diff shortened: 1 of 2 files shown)


mojavelinux:
Is there anything I need to do to get this PR accepted?

etagwerker (changes requested):
@mojavelinux Sorry for the delay. Could you please update the README with the results? Thanks!

etagwerker (approved):
Looks good, thanks! 💯

PR #124, opened 2017-05-05
Compare Hash#select#values vs Hash#values#select vs Hash#values#compact

sFrenkie:
This benchmark proves that get not nil values from hash is fastest with Hash#values#compact

### DIFF
--- a/code/hash/select-vs-values.compact.rb
+++ b/code/hash/select-vs-values.compact.rb
@@ -1,0 +1,24 @@
+require 'benchmark/ips'
+
+ARRAY = Array.new(1000) { Random.rand }
+VALUES = ARRAY.map { |v| v < 0.5 }
+HASH = Hash[ARRAY.zip(VALUES)]
+
+def fastest
+  HASH.values.compact
+end
+
+def fast
+  HASH.values.select { |v| v }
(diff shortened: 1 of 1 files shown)


etagwerker (changes requested):
I think this PR adds value if you keep the `fastest` and `fast` methods, but I feel like `slow` is doing something way different. 

`fast` and `fastest` are calling a method on `Array` (`compact` vs `select`) but `slow` is calling a method on `Hash` and it's a little too contrived.

etagwerker:
Thanks for this, @sFrenkie! The core idea is useful, so I've opened #236 which builds on it.

One thing to flag: in this version the three expressions aren't quite equivalent. The test data uses boolean values (`v < 0.5` → `true`/`false`) and contains no `nil` values, so `Hash#values.compact` filters nothing while the two `select` variants also drop `false`. That makes `compact` look fastest partly because it's doing less work.

#236 uses data that actually contains `nil` values so all three return the same result (plus an equivalence guard), and adds the README entry. The conclusion still holds: `Hash#values.compact` is the fastest way to get non-nil values. Closing this in favor of #236 — thanks again for the contribution!

PR #126, opened 2017-06-02
Add double splat operator vs merge

carlos-contreras:
Shows how the double splat operator adds faster Hash merging compared to the `Hash#merge` method.


Arcovion:
Thanks!

PR #127, opened 2017-06-07
Add benchmark for format() vs round().to_s

carlos-contreras:
This compares two alternatives for starting with a floating point
number, transform it into a formatted string with a specific number
of digits after the point.

### DIFF
--- a/README.md
+++ b/README.md
@@ -268,6 +268,27 @@ Comparison:
           OpenStruct:    96855.3 i/s - 16.56x slower
 ```
 
+##### Kernel#format vs Float#round().to_s [code](code/general/format-vs-round-and-to-s.rb)
+
+```
+$ ruby -v code/general/format-vs-round-and-t
+o-s.rb
+ruby 2.3.3p222 (2016-11-21 revision 56859) [x86_64-darwin15]
+Warming up --------------------------------------
+         Float#round   106.645k i/100ms
+       Kernel#format    84.304k i/100ms
+            String#%    78.635k i/100ms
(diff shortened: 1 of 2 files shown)


Arcovion:
Typo: `format('%.2f' % NUM)` &rarr; `format('%.2f', NUM)` or `'%.2f' % NUM`

Arcovion:
Strangely, on my machine `format` is consistently faster than `%`:
```ruby
require 'benchmark/ips'

NUM = 1.12678.freeze

def fast
  NUM.round(2).to_s
end

def avg
  format('%.2f', NUM)
end

def slow
  '%.2f' % NUM
end

Benchmark.ips do |x|
  x.report('Float#round') { fast }
  x.report('Kernel#format') { avg }
  x.report('String#%') { slow }
  x.compare!
end
```
```
         Float#round   128.526k i/100ms
       Kernel#format    96.841k i/100ms
            String#%    89.243k i/100ms
         Float#round      1.981M (± 0.3%) i/s -     10.025M in   5.061882s
       Kernel#format      1.505M (± 0.3%) i/s -      7.554M in   5.017527s
            String#%      1.241M (± 0.3%) i/s -      6.247M in   5.034723s

Comparison:
         Float#round:  1980506.7 i/s
       Kernel#format:  1505458.6 i/s - 1.32x  slower
            String#%:  1240793.9 i/s - 1.60x  slower
```
So might be worth including both in the benchmark

carlos-contreras:
Thanks for the feedback @Arcovion, I updated the PR 👍

PR #132, opened 2017-09-15
Add JSON.generate vs Oj.dump

carlos-contreras:
Compares two alternatives for generating a json format starting
from a base Hash, the result shows how the `oj` gem is 9X faster
than the default `json` gem.

### DIFF
--- a/README.md
+++ b/README.md
@@ -268,6 +268,24 @@ Comparison:
           OpenStruct:    96855.3 i/s - 16.56x slower
 ```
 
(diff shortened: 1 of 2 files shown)


carlos-contreras:
I don't get to be able to resolve the conflict locally, don't see any 🤔

stereobooster:
it is not always compatible. Plus it is MRI specific. But if we go this way there is also other gems dedicated to make MRI faster though C extensions like:

- gem "escape_utils"
- gem "fast_blank"
- gem "oj"
- gem "faster_path"
- gem "fast_gettext"
- gem "jemalloc"

PR #137, opened 2017-10-20
unintentional object allocation

stereobooster:
```
ruby code/method/default-arguments.rb
constant as default argument
                         0.000  memsize (     0.000  retained)
                         0.000  objects (     0.000  retained)
                         0.000  strings (     0.000  retained)
new object as default argument
                       400.000k memsize (     0.000  retained)
                        10.000k objects (     0.000  retained)
                         0.000  strings (     0.000  retained)

Comparison:
constant as default argument  :          0 allocated
new object as default argument:     400000 allocated - Infx more
```

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,5 +1,5 @@
 source 'https://rubygems.org'
 
 gem 'benchmark-ips', '>= 2.0'
-
+gem 'benchmark-memory'
 gem 'rake'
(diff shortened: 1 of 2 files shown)


benoittgt:
Related also : https://github.com/benoittgt/understand_ruby_memory#why-keywords-arguments-will-create-less-garbage-collected-objects-after-22-answered-

PR #138, opened 2017-11-01
Add String#to_sym vs colon

lvl0nax:


ixti:
That's like testing `"123".to_i` vs `123`.

PR #144, opened 2018-02-08
[Enhance] Add Another Inject-Like Method That Might Occur in Ruby for Beginners

berniechiu:
### Objective

Since some Ruby beginners still write aggregations without inject, it's good to show them a benchmark on different usage that might occur to them.

### Feature

- Updated ruby version with new benchmark
- Fix the changed file name

### DIFF
--- a/README.md
+++ b/README.md
@@ -595,21 +595,24 @@ Enumerable#sort_by (Symbol#to_proc):    25916.1 i/s
 Of note, `to_proc` for 1.8.7 is considerable slower than the block format
 
 ```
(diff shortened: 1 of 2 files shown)


berniechiu:
Wonder why some of the approved PRs not merging lol ~

ixti:
TBH IMO all that part should be completely rewritten. There's no point to refer rubies 1.x - they are EOL long time ago.

FunkyloverOne:
OMG, 30+x difference :o

PR #145, opened 2018-02-11
Remove extra spaces (or other contiguous chars)

styd:
The code is tested against contiguous spaces, but should work for other contiguous chars too.


ixti:
Thank you!

PR #153, opened 2018-04-08
Add Enumerable#sort_by.first vs Enumerable#min_by

parkerfinch:
As discussed in https://github.com/bbatsov/rubocop/pull/5753


Arcovion:
Thanks!

PR #163, opened 2018-08-26
Update `Proc#call` and block arguments vs `yield` for Ruby 2.5 and 2.6

bdewater:
In Ruby 2.5, passing through a `block` argument has become a lot faster.

I renamed the 'block argument' benchmark because I felt the old name was misleading as the fastest on Ruby 2.5.


bdewater:
I've updated the PR to include Ruby 2.6.

Arcovion:
Thanks!

PR #168, opened 2019-02-01
Add Array#reverse.first vs Array#last.reverse and vice-versa.

Oppen:


ashmaroli:
@Oppen You've got the methods reversed in the benchmark script. The label says one thing but the method does the opposite. Please correct!

Also, it would be clearer to others if the labels stated that this relates to capturing multiple elements.
e.g. `'Array#last(n > 1).reverse'` instead of `'Array#last.reverse'`
The latter label implies that you're calling `.reverse` on the last item of the Array

Oppen:
I'll fix the benchmark when I have time.
While it's true the label could be more informative, note it actually
applies to the last element, too, so it would be `n >= 1`.

ashmaroli:
> note it actually applies to the last element

@Oppen Consider the following:
```ruby
array = %w(alpha beta gamma delta)
p array.last.reverse
p array.reverse.first
```
Are they the same?

```ruby
array = %w(alpha beta gamma delta)
p array.last(2).reverse
p array.reverse.first(2)
```
Now are they the same?

Oppen:
Right. I didn't realize the types would change. Thanks!

Oppen:
However, I run a simpler test:
["alpha"].first(1)
And the result was ["alpha"], so it actually is a matter of whether you
pass a parameter or not, not about the number of elements.
I'm not sure how I should express that in the labels, but that's useful
mostly because n may come as a parameter to the caller.

ashmaroli:
> I'm not sure how I should express that in the labels

I think taking a cue from your sample method is sufficient.
Either ways, thank you for reporting this.

```ruby
Benchmark.ips do |x|
  x.report('Array.last(n).reverse')  { ARRAY.last(5).reverse }
  x.report('Array.reverse.first(n)') { ARRAY.reverse.first(5) }
  x.compare!
end
```

PR #173, opened 2019-02-24
Native Hash#slice vs other slice alternatives

styd:
Since ruby v2.5, Hash comes with `slice` method to select members by keys.


code/hash/slice-vs-select-include.rb
```diff
@@ -0,0 +1,22 @@
...
+  author: "styd",
+  published_at: Time.now
+}
+
+def fast
+  HASH.slice(:title, :author)
+end
+
+def slow
+  HASH.select {|k, _| [:title, :author].include? k }
```
bdewater:
```
[:title, :author].each_with_object(Hash.new) { |k, h| h[k] = HASH[k] if h.has_key?(k) }
```
Based on [Active Support](https://github.com/rails/rails/blob/94b5cd3a20edadd6f6b8cf0bdf1a4d4919df86cb/activesupport/lib/active_support/core_ext/hash/slice.rb#L23-L25), which is a bit faster than this on my machine... but I don't think comparing native methods with their Ruby equivalent is really the point, of course that'll be faster.

styd: You are right. That's what I thought before making this PR.
But I PR anyway because:
1. To show other rubyist who doesn't already know that they have another alternative in newer ruby.
2. To show other rubyist who already know the `slice` method that it is not just a method that wraps the ruby select-includes.

ixti: @styd Your `slow` implementation is basically simply bad code. You're creating an array on each key/par cycle. So it worth to replace `slow` method with something good ruby code would look like (pretty much what @bdewater provided).

Arcovion:
```ruby
require 'benchmark/ips'

HASH = {
  title: "awesome",
  description: "a description",
  author: "styd",
  published_at: Time.now
}

HASH.dup.each{ |k, v| HASH[:"#{k}_2"] = v }

def fast
  HASH.slice(*%i[title author other])
end

def rails
  %i[title author other].each_with_object({}){ |k, h| h[k] = HASH[k] if HASH.key?(k) }
end

def slow_with_few_keys
  (%i[title author other] & HASH.keys).to_h{ |k| HASH.assoc(k) }
end

def slow_with_many_keys
  HASH.select{ |k, _| %i[title author other].include?(k) }
end

Benchmark.ips do |x|
  x.report('fast') { fast }
  x.report('rails') { rails }
  x.report('slow_with_few_keys') { slow_with_few_keys }
  x.report('slow_with_many_keys') { slow_with_many_keys }
  x.compare!
end
```
```
Comparison:
                fast:   652730.7 i/s
               rails:   293760.7 i/s - 2.22x  slower
  slow_with_few_keys:   242545.9 i/s - 2.69x  slower
 slow_with_many_keys:   166966.4 i/s - 3.91x  slower
```
I tried a bunch of other methods and the rails way was always the fastest except for `slice`.
The [first result for 'ruby filter hash' or 'ruby select hash'](https://stackoverflow.com/a/7430444) mentions `slice` was added in Ruby 2.5. 
Rails users will also have known about it for a while.
Closing as I think most people will search for a more elegant way on their own and realise they can't do better than `slice`.

ixti:
@Arcovion Using `each` instead of `each_with_object` will be faster:

``` ruby
require "benchmark/ips"

HASH = {
  title: "awesome",
  description: "a description",
  author: "styd",
  published_at: Time.now
}

HASH.dup.each{ |k, v| HASH[:"#{k}_2"] = v }

def native
  HASH.slice(*%i[title author other])
end

def ruby
  memo = {}
  %i[title author other].each { |k| memo[k] = HASH[k] if HASH.key?(k) }
  memo
end

def rails
  %i[title author other].each_with_object({}){ |k, h| h[k] = HASH[k] if HASH.key?(k) }
end

puts "RUBY: #{RUBY_VERSION}"

Benchmark.ips do |x|
  x.report("native") { native }
  x.report("ruby") { ruby }
  x.report("rails") { rails }
  x.compare!
end
```

---

```
RUBY: 2.6.1
              native   308.802k i/100ms
                ruby   189.104k i/100ms
               rails   161.126k i/100ms
              native      5.384M (± 3.5%) i/s -     27.175M in   5.054550s
                ruby      2.605M (± 2.1%) i/s -     13.048M in   5.011819s
               rails      2.062M (± 2.6%) i/s -     10.312M in   5.004325s

Comparison:
              native:  5384259.3 i/s
                ruby:  2604735.1 i/s - 2.07x  slower
               rails:  2062111.3 i/s - 2.61x  slower
```

styd:
This is good stuff. I'm glad I did this PR.

@Arcovion If the reason for closing this PR is because people already/eventually know that `slice` is better, why would there be `Array#shuffle.first vs Array#sample`. Shouldn't people eventually know too that `sample` is better.
If you think adding `Rails`' version is important, why won't we do that?
We can also add @ixti's version that is better than `Rails`'.

I really believe `slice` needs more exposure just as `sample` was and one way of doing that is putting it in this repo.

Arcovion:
Good point, to me the examples where you combine two methods into one are 'patterns' you can easily spot in every day code. This feels more involved, where people could come up with multiple ways to do it, but if we try to include all those different ways  here then it might be good to have.

styd:
@Arcovion Noted. Thanks for reopening. Please review again.

PR #176, opened 2019-05-26
Add Array#sort.reverse vs Array#sort_by with block

sshaw:


semaperepelitsa:
Should really be compared with `sort_by{ |a| -a }`or `sort_by(&:-@)`

ixti:
`arr.reverse.sort` simply makes no sense. `Array#reverse` returns copy of array with reversed order.

Comparing `sort_by { |a| -a }` to `sort_by(&:-@)` was already covered in:

https://github.com/JuanitoFatas/fast-ruby/blob/master/code/enumerable/sort-vs-sort_by.rb

Comparing `sort { |a, b| b <=> a }` to `sort_by(&:-@)` is also pointless as it's already covered by the above.

ParadoxV5:
Did you mean, `sort.reverse!`?

sshaw:
> Should really be compared with sort_by{ |a| -a }or sort_by(&:-@)

@semaperepelitsa agreed. Updated. 

> arr.reverse.sort simply makes no sense.

@ixti typo, should be `arr.sort.reverse`. Fixed. 

> Comparing sort_by { |a| -a } to sort_by(&:-@) was already covered in:

@ixti yes I know, I wrote it. This PR differs by 50%. 

> Did you mean, sort.reverse!?

@ParadoxV5 yes updated.

sshaw:
Though technically this is comparing `Array#sort` etc... not `Enumerable`. Updated.

ParadoxV5:
> though no `!`
@sshaw Not sure if the `sort!` variant is faster or not because several other languages (including our rival Python) do their `sort`ing in-place (modifying the original). (`reverse` and `sort` each creates a new array with reversed contents.)

Also, should compare with `_2 <=> _1` because `Numeric` happens to have negation functionality (`@-`), but not guaranteed any ~~`Comparable`~~ __duck-type with `<=>`__, such as `String`s (`@-` doesn’t negate) and `Array`s (`@-` isn’t defined).

P.S. @ixti The above file does not sort `reverse`dly.

ixti:
I still think this comparison is weird and very niche - it has meaning only when the object is an `Array` of `Number`s. In which case the bet variant will be: `arr.sort.tap(&:reverse!)`, in other cases it will either fail or will give unexpected result.

@ParadoxV5 I know that the `sort-vs-sort_by.rb` has no reverse sorting.

etagwerker (approved):
@sshaw Thanks for submitting this! I think it adds value to the project.

PR #178, opened 2019-12-28
Update some examples with Ruby 2.7

RicardoTrindade:
Now that Ruby 2.7 is out I've updated the date.parse vs date.iso8601 example, strings tr vs gsub and mutable vs immutable strings example.
Happy to update more examples if this one goes forward.
Ran on a MacBook Pro (Retina, 15-inch, Mid 2015), 2.2 GHz Intel Core i7, 16 GB 1600 MHz DDR3

### DIFF
--- a/README.md
+++ b/README.md
@@ -663,17 +663,17 @@ When expecting well-formatted data from e.g. an API, `iso8601` is faster and wil
 
 ```
 $ ruby -v code/date/iso8601-vs-parse.rb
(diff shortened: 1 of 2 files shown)


ixti:
I think we should automate that instead of manual write-ups.

nirvdrum:
Indeed. It'd be helpful if these numbers were generated on the same hardware, too. Otherwise, there's just going to be too much variation from run-to-run.

RicardoTrindade:
Understand your points, will close it then.

PR #183, opened 2020-06-09
Add Array#to_h comparison with several other implementations

sebastian-palma:
As Ruby 2.6 added support for executing a block when passed to `to_h` I think we can add it with the other implementations to achieve the same output.

Ruby on Rails also has a shorter way [`Enumerable#index_by`](https://api.rubyonrails.org/classes/Enumerable.html#method-i-index_by) which I didn't want to include, but it's worth mentioning for those working with the framework.

### DIFF
--- a/code/array/toh-vs-each-vs-Hash[]-map-vs-map-toh-vs-each-with-object-vs-tap-each.rb
+++ b/code/array/toh-vs-each-vs-Hash[]-map-vs-map-toh-vs-each-with-object-vs-tap-each.rb
@@ -1,0 +1,56 @@
+# frozen_string_literal: true
+
+require "benchmark/ips"
+
+class People
+  def initialize(first_name, last_name)
+    @first_name = first_name
+    @last_name = last_name
+  end
+
+  def login
+    "#{first_name.downcase}-#{last_name.downcase}"
+  end
+
+  private
+
+  attr_reader :first_name, :last_name
+end
+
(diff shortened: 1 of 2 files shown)


code/array/toh-vs-each-vs-Hash[]-map-vs-map-toh-vs-each-with-object-vs-tap-each.rb
```diff
@@ -0,0 +1,56 @@
...
+
+  private
+
+  attr_reader :first_name, :last_name
+end
+
+PEOPLE = [People.new('John', 'Doe'), People.new('Lola', 'Lanos')].freeze
+
+def fast
+  PEOPLE.to_h { |person| [person.login, person] }
```
Comment (nirvdrum) [inline code/array/toh-vs-each-vs-Hash[]-map-vs-map-toh-vs-each-with-object-vs-tap-each.rb:23]: To be fair to the other benchmarks, this should also call `dup` on `PEOPLE`, I think.

Comment (sebastian-palma) [reply code/array/toh-vs-each-vs-Hash[]-map-vs-map-toh-vs-each-with-object-vs-tap-each.rb:23]: Oh yes @nirvdrum, thanks! It might look suspicious without it.

Comment (nirvdrum) [reply code/array/toh-vs-each-vs-Hash[]-map-vs-map-toh-vs-each-with-object-vs-tap-each.rb]: If every other benchmark calls `dup`, you're not isolating your changes to the various "to hash" methods. This one will look faster because it's not copying objects. I suppose that's fair in highly niche cases, but for the most part, I don't think people are calling `to_h` on a frozen constant array.

PR #184, opened 2020-06-21
ParadoxV5’s additions

ParadoxV5:
I would like to contribute. I hope I didn’t necro-“thread”!
```
Rational() vs divide with r-literals to create rationals whose decimals are infinitely repeating, and `if block_given?` vs `if block`.
Comparision tested with regular Benchmark with own computer (Ruby 2.7.0) and repl.it (Ruby 2.5.?)
```
I might have more commits/PRs coming up.

### DIFF
--- a/code/general/rational-vs-divide-with-r.rb
+++ b/code/general/rational-vs-divide-with-r.rb
@@ -1,0 +1,26 @@
+def avg
+  Rational(2, 3)
+end
+def slow1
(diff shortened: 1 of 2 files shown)


ParadoxV5:
Here, I believe these are all my ideas (excluding ones with not much speed difference) for now.
Go ahead and review my submissions.
🧡

PR #193, opened 2021-06-10
Hash#has_key? vs Set#include?

marcotc:
When selecting a data structure to store a set of unique elements, in which a common operation will be to check if an element is part of that data structure, the a [`Set`](https://ruby-doc.org/stdlib-3.0.1/libdoc/set/rdoc/Set.html) comes to mind as the ideal, minimal choice.

One downside of `Set` is that it's slower than a `Hash` when checking for present of an element, even though `Set`'s implementation is [backed by a `Hash`](https://github.com/ruby/ruby/blob/929cc615a749f467809a865a3d40adcc0b58c667/lib/set.rb#L401-L403).

This PR documents the difference in performance profile between `Hash#has_key?` and `Set#include?` to help users that need to check for an association in performance critical paths.

### DIFF
--- a/code/hash/set-vs-hash-include.rb
+++ b/code/hash/set-vs-hash-include.rb
@@ -1,0 +1,19 @@
+require 'benchmark/ips'
+require 'set'
+
+HASH = Hash[*('a'..'z').map { |key| [key, true] }]
+SET = Set[*('a'..'z').to_a]
+
+def fast
+  HASH.has_key?('n')
+end
+
+def slow
+  SET.include?('n')
+end
+
+Benchmark.ips do |x|
+  x.report('Hash#has_key?') { fast }
+  x.report('Set#include?') { slow }
+  x.compare!
+end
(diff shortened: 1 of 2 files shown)


code/hash/set-vs-hash-include.rb
```diff
@@ -0,0 +1,19 @@
+require 'benchmark/ips'
+require 'set'
+
+HASH = Hash[*('a'..'z').map { |key| [key, true] }]
```
krtschmr, line 4: this seems very weird to me. are you sure you want to create a hash like this?
```
 {["a", true]=>["b", true],
 ["c", true]=>["d", true],
 ["e", true]=>["f", true],
 ["g", true]=>["h", true],
 ["i", true]=>["j", true],
 ["k", true]=>["l", true],
 ["m", true]=>["n", true],
 ["o", true]=>["p", true],
 ["q", true]=>["r", true],
 ["s", true]=>["t", true],
 ["u", true]=>["v", true],
 ["w", true]=>["x", true],
 ["y", true]=>["z", true]}
```

etagwerker:
A benchmark should use the same starting point. This benchmark is using two different objects. I don't think this PR is a good fit for this project.

Issue #194, opened 2021-08-06
Fast Ruby for hash_update_vs_hash_brackets


manjunath724:
**Hash.update** is an alias for **Hash.merge!**

Currently we have `merge_bang_vs_[]=` which suggests that  **Hash#[]=** is faster and **Hash#merge!()** is slower.
But we dont have any suggestions for `update_vs_[]=` i.e., **Hash#update()** vs **Hash#[]=**.

```
require 'benchmark/ips'

ENUM = (1..100)

def fast
  ENUM.each_with_object({}) do |e, h|
    h[e] = e
  end
end

def slow
  ENUM.each_with_object({}) do |e, h|
    h.update(e => e)
  end
end

Benchmark.ips do |x|
  x.report('Hash#[]=') { fast }
  x.report('Hash#update') { slow }
  x.compare!
end
```

```
##### `Hash#update` vs `Hash#[]=` [code](code/hash/update-vs-\[\]=.rb)

$ ruby -v code/hash/update-vs-\[\]=.rb
ruby 2.6.6p146 (2020-03-31 revision 67876) [x86_64-darwin18]

Warming up --------------------------------------
            Hash#[]=     7.453k i/100ms
         Hash#update     4.311k i/100ms
Calculating -------------------------------------
            Hash#[]=     74.764k (± 1.9%) i/s -    380.103k in   5.085962s
         Hash#update     43.220k (± 0.8%) i/s -    219.861k in   5.087364s

Comparison:
            Hash#[]=:    74764.0 i/s
         Hash#update:    43220.1 i/s - 1.73x  (± 0.00) slower
```

Can we extend this to Hash.update()?

PR #195, opened 2021-08-06
Hash#update() vs Hash#[]=

manjunath724:
**Hash#update** is an alias for **Hash#merge!**. Ergo, the comparision between **Hash#update** and **Hash#[]=** should have similar outcome.

Ref https://github.com/JuanitoFatas/fast-ruby/issues/194

### DIFF
--- a/code/hash/update-vs-[]=.rb
+++ b/code/hash/update-vs-[]=.rb
@@ -1,0 +1,21 @@
+require 'benchmark/ips'
+
+ENUM = (1..100)
+
+def fast
+  ENUM.each_with_object({}) do |e, h|
+    h[e] = e
+  end
+end
+
+def slow
+  ENUM.each_with_object({}) do |e, h|
+    h.update(e => e)
+  end
+end
+
+Benchmark.ips do |x|
+  x.report('Hash#[]=') { fast }
+  x.report('Hash#update') { slow }
+  x.compare!
+end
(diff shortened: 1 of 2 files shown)


code/hash/update-vs-[]=.rb
```diff
@@ -0,0 +1,21 @@
...
+
+def fast
+  ENUM.each_with_object({}) do |e, h|
+    h[e] = e
+  end
+end
+
+def slow
+  ENUM.each_with_object({}) do |e, h|
+    h.update(e => e)
```
Comment (mblumtritt) [inline code/hash/update-vs-[]=.rb:13]: Here you create an additional Hash object which of curse will be much slower than only provide two "static" parameters as for `Hash#[]`.

Please, only compare things which are comparable.

Comment (manjunath724) [reply code/hash/update-vs-[]=.rb:13]: Agree but keeping it similar to its alias [`Hash#merge!`](https://github.com/JuanitoFatas/fast-ruby/blob/master/code/hash/merge-bang-vs-%5B%5D%3D.rb)

Comment (manjunath724) [reply code/hash/update-vs-[]=.rb:13]: Can you please guide me, how to go about it?

Comment (manjunath724) [reply code/hash/update-vs-[]=.rb:13]: Thank you @krtschmr As a next step can the Travis CI build be run for this PR and if everything looks good could this PR be considered for next release?

PR #198, opened 2021-12-02
Compare one n-element hash#merge! vs n assignments with []=

ohaleck:
The [code/hash/merge-bang-vs-[]=.rb](https://github.com/JuanitoFatas/fast-ruby/blob/38f49f95fc7574d929de60b71791d09129c2588c/code/hash/merge-bang-vs-%5B%5D=.rb) benchmark merges N 1-element hashes to compare it with N hash element assignments using `[]=`. Obviously assignments are faster, but this is not a real-life scenario.

The more common case would be to perform one merge an N-element hash and it appears it is much faster to do it than to perform N assignments. This PR implements such a benchmark.

### DIFF
--- a/code/hash/n-merge-bang-vs-[]=.rb
+++ b/code/hash/n-merge-bang-vs-[]=.rb
@@ -1,0 +1,22 @@
+require 'benchmark/ips'
+
+ENUM = (1..100)
+HASH = ENUM.to_h{ |a| [a, a]}
+SECOND_HASH = ENUM.to_h{ |a| [a+50, a+50]}
+
+def merge
+  h = HASH
+  h.merge!(SECOND_HASH)
+end
+
+def assign
+  ENUM.each_with_object({}) do |e, h|
+    h[e] = e
+  end
+end
+
+Benchmark.ips do |x|
+  x.report('Hash#merge!') { merge }
+  x.report('Hash#[]=') { assign }
+  x.compare!
+end


code/hash/n-merge-bang-vs-[]=.rb
```diff
@@ -0,0 +1,22 @@
+require 'benchmark/ips'
+
+ENUM = (1..100)
+HASH = ENUM.to_h{ |a| [a, a]}
+SECOND_HASH = ENUM.to_h{ |a| [a+50, a+50]}
+
+def merge
+  h = HASH
+  h.merge!(SECOND_HASH)
```
Comment (mblumtritt) [inline code/hash/n-merge-bang-vs-[]=.rb:9]: This changes the (not longer) constant `HASH` and that seem to be intended. You can just use `HASH#merge`.

Comment (ParadoxV5) [reply code/hash/n-merge-bang-vs-[]=.rb:9]: `[]=` also changes `HASH`

Comment (mblumtritt) [reply code/hash/n-merge-bang-vs-[]=.rb:9]: @ParadoxV5 ?!
Do you mean the assignment to `h` below? In this case `h` is created in line 13…

code/hash/n-merge-bang-vs-[]=.rb
```diff
@@ -0,0 +1,22 @@
...
+end
+
+def assign
+  ENUM.each_with_object({}) do |e, h|
```
Comment (ParadoxV5) [inline code/hash/n-merge-bang-vs-[]=.rb:13]: Function `merge` works with a hash prefilled with 100 entries, while `assign` works with a blank new hash.
```suggestion
  ENUM.each_with_object(SECOND_HASH) do |e, h|
```

Comment (ParadoxV5) [reply code/hash/n-merge-bang-vs-[]=.rb:13]: P.S. though my suggestion is [`ENUM` ➡ `SECOND_HASH`] while `merge` is [`SECOND_HASH` ➡ hash version of `ENUM`]

etagwerker:
Closing due to inactivity. If anyone is interested in these changes, please submit a new PR addressing all the comments in this PR. Thanks! ❤️

PR #203, opened 2022-07-07
Migrate from Travis CI to GitHub Actions 

### LINKED ISSUE #202: [FEATURE REQUEST] Run all the benchmarks on CI
We could use github actions and a matrix to run the benchmarks on CI and with different rubies. That way we could compare what the readme shows with what CI shows and compare results between different rubies too for the same benchmark since some of these things may change if ruby optimizes something.

There is a travis configuration but the link is broken. We could use GitHub Actions instead.

etagwerker:
Hi there,

This PR fixes #202 and adds sharing as a default for the rake task. Sharing is set up to only run in CI, so that it doesn't run on development environments when people are trying to make a contribution to this project... 

Please check it out.

Thanks!


ixti (approved):

arielj (approved):
Looks good to me.

Just a suggestion but shouldn't be a blocker: maybe rename the workflow to something like `Benchmarks` and `benchmarks.yml`, currently it's `Ruby CI` an the filename is `rake.yml`.

One question that might be outside of the scope of this PR, I see the benchmarks are shared at ips.fastruby.io, mabye this should happen only if the tests are run on the main branch? I imagine sharing results for every PR change might be a lot, but I don't know how ips.fastruby.io work (maybe this behavoir is intentional, just asking)

arielj:
One more thing I forgot, can you update the README file removing the reference to Travis CI? Maybe link to the Actions index.

etagwerker:
@arielj Just merged this PR and added a new issue to address the sharing enhancement: https://github.com/fastruby/fast-ruby/issues/207 👍

Issue #53, opened 2015-08-05
ends_with? advice incorrect


willbryant:
The following text in the regex vs. starts_with?/ends_with? bit (which has a ! next to it) is flatly incorrect:

```
You can combine start_with? and end_with? to replace error.path =~ /^#{path}(\.rb)?$/ to this 
error.path.start_with?(path) && error.path.end_with?('.rb', '')
—— @igas rails/rails#17316
```

`error.path.end_with?('.rb', ‘’)` is equivalent to `error.path.end_with?(‘’)` which always returns true and is therefore useless.  In addition, the combination of starts_with? and ends_with? does not test that nothing else is between `path` and the end except `.rb` like the regexp does.

(Notably the cited PR was rejected and it was noted that some of the other suggested replacements were wrong.)

This paragraph should be just deleted, I think.

PR #211, opened 2022-10-04
Removes wrong advice concerning `end_with?`

mateusdeap:
Solves #53


JuanVqz (approved):
This makes sense, thanks @mateusdeap!

PR #214, opened 2022-11-05
Run benchmarks on TruffleRuby in CI.

nirvdrum:
TruffleRuby is a high speed implementation of Ruby. It should run all of these benchmarks just fine. I think it'd be interesting to see the results. I suspect in many cases, the JIT can optimize the different variants of a given benchmark so that there is no clear winner. In cases where it does not, that indicates an opportunity for the JIT to improve.


etagwerker (approved):
Looks good, thanks!

PR #216, opened 2022-11-12
Randomized string concat

alekseyl:
proposition for https://github.com/fastruby/fast-ruby/issues/158 fix.

### DIFF
--- a/code/string/concatenation_randomized.rb
+++ b/code/string/concatenation_randomized.rb
@@ -28,8 +28,8 @@ module RandStr
   def self.method_missing(symbol)
     return super unless @rand_strs.keys.include?(symbol)
 
-    define_singleton_method(symbol) { rand_str( symbol ) }
-    return rand_str( symbol )
+    define_singleton_method(symbol) { rand_str(symbol) }
+    return rand_str(symbol)
   end
 
 end
--- a/code/string/concatenation_randomized.rb
+++ b/code/string/concatenation_randomized.rb
@@ -13,8 +13,8 @@ module RandStr
   def self.generate_rand_strs
     chars = ('A'..'z').to_a * 20
     @rand_strs[:lt10] = Array.new(RND_STRINGS_AMOUNT) { chars.sample(rand(10)).join }
-    @rand_strs[:lt100] = Array.new(RND_STRINGS_AMOUNT) { chars.sample( rand(100) ).join }
-    @rand_strs[:lt1000] = Array.new(RND_STRINGS_AMOUNT) { chars.sample( rand(1000) ).join }
+    @rand_strs[:lt100] = Array.new(RND_STRINGS_AMOUNT) { chars.sample(rand(100)).join }
+    @rand_strs[:lt1000] = Array.new(RND_STRINGS_AMOUNT) { chars.sample(rand(1000)).join }
     @rand_strs[:eq10] = Array.new(RND_STRINGS_AMOUNT) { chars.sample(10).join }
     @rand_strs[:eq100] = Array.new(RND_STRINGS_AMOUNT) { chars.sample(100).join }
   end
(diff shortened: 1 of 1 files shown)


JuanVqz (commented):
Hey, @alekseyl thank you for the time you spent on this, it looks good, I just left some minor comments, let me know what you think :+1:

alekseyl:
Thnx @JuanVqz, no worries for the delay :). I've commited suggested!

code/string/concatenation_randomized.rb
```diff
@@ -0,0 +1,155 @@
...
+      sarr2 = Array.new(2_000_000) { RandStr.eq100.dup }
+  
+      i, j = 0, 0
+      x.report("Collateral actions only")  { k=0; 1_000_000.times { k+=1; RandStr.eq10; sarr2[k]; RandStr.lt100; } }
+      x.report("String#+")         { k=0; 1_000_000.times { k+=1; sarr1[k]; fastest_plus(RandStr.eq10, RandStr.lt100) } }
+      x.report('"#{foo}#{bar}"')   { k=0; 1_000_000.times { k+=1; sarr2[k]; fast_interpolation(RandStr.eq10, RandStr.lt100) } }
+      x.report("String#concat")    { 1_000_000.times { RandStr.eq10; slow_concat(sarr1[i], RandStr.lt100); i+=1; } }
+      x.report("String#append")    { 1_000_000.times { RandStr.eq10; slow_append(sarr2[j], RandStr.lt100); j+=1;  } }
+    end
+end
```
JuanVqz, line 154:
```suggestion
def bench_10_to_100
  Benchmark.bmbm do |x|
    sarr1 = Array.new(2_000_000) { RandStr.eq100.dup }
    sarr2 = Array.new(2_000_000) { RandStr.eq100.dup }
  
    i, j = 0, 0
    x.report("Collateral actions only")  { k=0; 1_000_000.times { k+=1; RandStr.eq10; sarr2[k]; RandStr.lt100; } }
    x.report("String#+")         { k=0; 1_000_000.times { k+=1; sarr1[k]; fastest_plus(RandStr.eq10, RandStr.lt100) } }
    x.report('"#{foo}#{bar}"')   { k=0; 1_000_000.times { k+=1; sarr2[k]; fast_interpolation(RandStr.eq10, RandStr.lt100) } }
    x.report("String#concat")    { 1_000_000.times { RandStr.eq10; slow_concat(sarr1[i], RandStr.lt100); i+=1; } }
    x.report("String#append")    { 1_000_000.times { RandStr.eq10; slow_append(sarr2[j], RandStr.lt100); j+=1;  } }
  end
end
```

JuanVqz (approved):
@alekseyl, Thank you for your contribution, it looks good :+1:

JuanVqz:
@mateusdeap could you please have a look to merge it?

etagwerker (changes requested):
@alekseyl @JuanVqz I noticed that this PR uses `bmbm` and we want all benchmarks to use `ips`

Could you please tweak this benchmark to use "benchmark/ips"? 

It shouldn't affect the results. 

Thanks!

alekseyl:
> @alekseyl @JuanVqz I noticed that this PR uses bmbm and we want all benchmarks to use ips
> Could you please tweak this benchmark to use "benchmark/ips"?
> It shouldn't affect the results.
> Thanks!

@etagwerker Hi Ernesto!

Sorry for the delayed answer. The problem is it will not look as good as we migh expect, because of a randomization )). It will show a much bigger variance (20-30-40% I can't recall properly) over relatively "small" difference. 

I might give it a try one more time, but can't promise that it will be soon enough :(. I need an inspiration to dive in to that river again  ))).

ixti:
Can we also add:

``` ruby
[foo, bar].join
```

It would also make sense to check if there will be different results with more than 2 strings to concatenate. Also, would be nice to see performance comparison of cases like:

``` ruby
"#{foo}-#{bar}"
```

etagwerker (approved):
@alekseyl Looks good, thank you!

PR #223, opened 2023-11-04
CI: Generate report file artifacts for each ruby version

ishigo974:


ishigo974:
Sorry, misclick, still working on it.

PR #228, opened 2023-12-10
Add more examples to infinite loops

ydakuka:


code/general/loop-vs-while-true-vs-until-false-vs-infinite-range.rb
```diff
@@ -0,0 +1,42 @@
...
+  loop do
+    break if index > NUMBER
+    index += 1
+  end
+end
+
+def slowest
+  (0..).each do |index|
+    break if index > NUMBER
+    index += 1
```
mblumtritt, line 32:
```suggestion
```

The index is already incremeted (you break too early at `Number / 2`)!

ydakuka: Oops. I've fixed it. Thanks.

PR #234, opened 2025-12-28
Update all benchmarks in the README to Ruby 4.0.0

kewlar:
As it's been 10 years since Ruby 2.2 was released, it may be a good time to re-run the benchmarks and see if the recommendations still stand.

(Actually very few surprises there.)

I also took the liberty of deleting the comparison between Rubies 2.4, 2.5, and 2.6 for `code/proc-and-block/block-vs-to_proc.rb`, as it can probably be safely considered as old news.


etagwerker (approved):
This is great, thank you! 🙏

